PR #19, opened 2024-06-22
Fix Installation of Nvidia Container Toolkit in `dev_setup.sh`

Zolkin1:
Attempted the install with `dev_setup.sh` on my laptop (fresh Ubuntu 22 install), and got errors with the Nvidia Container Toolkit install. Looks like these lines were different from the original Nvidia docs. When I ran these it seemed to work. Updated the setup file to reflect this.


alberthli (approved):
lgtm, added `-y` originally because I think there may have been options it asked about originally, but if we expect the user to be at their computer ready to respond to prompts from the install, this might be fine.

Note that if we want to eventually add `dev_setup.sh` as an automatically run script in a Docker container, it's very important that there's an exposed option to automatically accept user inputs, or the Docker container will not build.

PR #20, opened 2024-06-22
Dummy Bot: Obelisk Library Python Files

alberthli:
This PR addresses the python implementation of the first leg of the Dummy Bot features:

- [x] Other dirty QoL changes that I'm lumping into this PR since the codebase is already very unstable and we're trying to dev quickly
    - added Live Server and CMake Tools extensions to the dev container
    - added a `workaround` pixi task for downgrading `pytest` in order to circumvent bug where Humble from RoboStack and `pytest` 8.2.2 are incompatible
    - enforce Google python docstring conventions
    - allow more than 5 arguments in function definitions
    - fixed small bug in `source-obelisk` task where incorrect directory was sourced. also after appending the source command to `~/.bashrc`, actually also run source in the current terminal so no terminal restart is required.
    - deleted script that underlays original ROS2 (I had issues with finding it in my filesystem, I think something about RoboStack makes it so that you don't need to source it?)
- [x] Create empty message types as discussed face-to-face to expose the message API but without implementing it (will implement message fields in later PR)
- [x] Creates an `ObeliskNode` whose publishers/subscribers can only publish or subscribe to messages that we classify as Obelisk messages (with a finite number of exceptions, like `ParameterEvent`)
    - to facilitate this, I defined `OBELISK_MSG_TYPES` and `ROS_ALLOWED_MSG_TYPES` as variables that can be imported from `obelisk_py`. They contain respectively all custom obelisk messages we define as well as messages that are associated with default publishers on the ROS2 side (see `obelisk_py/__init__.py`).
    -  I defined a custom `ObeliskMsgError` for semantic clarity denoting the situation where users try to pass an unsupported message type into a publisher or subscriber.
    - wrote a util to get all the classes associated with any module. This is how we get all the message type automatically for each obelisk_msg subpackage.

Because this PR is large, in a second follow-up PR, I will address the following TODO:
- Create the abstract interfaces for the `Controller`, `WorldInterface`, `SimInterface`, and `StateEstimator` blocks of the Obelisk standard block diagram


obelisk_ws/src/obelisk_msgs/obelisk_control_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.5)
```
Zolkin1, line 1: In general its hard to say exactly what we should set this too. 3.0 and greater are very different, so we want at least that (which it is). That being said, I think we can get away with higher than 3.5. I have been using min 3.16 in my projects recently, but there isn't necessarily going to be any notable difference. I think we should bump this to 3.16 though, as that is what we have in the pixi.toml.

obelisk_ws/src/obelisk_msgs/obelisk_control_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+project(obelisk_control_msgs)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
```
Zolkin1, line 7: Unless ros requires a lower standard, I think we should set this to C++ 17 - there are some nice things in 17 I think we will use (like the filesystem library iirc)

obelisk_ws/src/obelisk_msgs/obelisk_estimator_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+set(THIS_PACKAGE_INCLUDE_DEPENDS
+  ament_cmake
+  std_msgs
+  rosidl_default_generators
+)
+include_directories(${THIS_PACKAGE_INCLUDE_DEPENDS})
+
+ament_export_dependencies(rosidl_default_runtime)
+ament_package()
```
Zolkin1, line 31: See above comments on versions.

obelisk_ws/src/obelisk_msgs/obelisk_sensor_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+
+set(THIS_PACKAGE_INCLUDE_DEPENDS
+  ament_cmake
+  std_msgs
+  rosidl_default_generators
+)
+include_directories(${THIS_PACKAGE_INCLUDE_DEPENDS})
+
+ament_export_dependencies(rosidl_default_runtime)
+ament_package()
```
Zolkin1, line 33: See above comments on versions

Zolkin1 (changes requested):
lgtm minus some of the cmake versioning stuff

obelisk_ws/src/obelisk_msgs/obelisk_control_msgs/CMakeLists.txt
alberthli: done.

obelisk_ws/src/obelisk_msgs/obelisk_control_msgs/CMakeLists.txt
alberthli: done.

Zolkin1 (approved):

PR #21, opened 2024-06-22
Add additional `obelisk_py` generic classes

alberthli:
Continues the work of #20.

Additional changes besides main tasks:

- new `obelisk_typing.py` file created to store all relevant types for type checking and dynamic typing use-cases like in `ObeliskNode`. for consistency, all types are now specified as `TypeVar`s to prevent ambiguity about what fields the types may have.
- major upgrades to `ObeliskNode`. Now, the core idea for configuring these nodes is that callback groups, timers, publishers, and subscribers can be configured by what is called a configuration string. This is a stringified version of the information contained in some configuration yaml file that the user supplies to launch files to bring up Obelisk nodes. This massively simplifies the creation of expected publisher/subscriber combinations in each type of node, like controllers, estimators, etc.
- allow user to manually allow non-Obelisk publishers/subscribers if they really want to for maximum flexibility
- removed `workaround` script in `pixi.toml` for downgrading `pytest`. Easier to just specify the `pytest` version as a non-pypi dependency.

Obelisk Modules
- [x] ObeliskController
- [x] ObeliskEstimator
- [x] ObeliskSensor
- [x] ObeliskRobot
- [x] ObeliskSim(ObeliskRobot)

Maintenance/Code
- [x] Separate tests into semantically different test functions cleanly - currently pretty messy
- [x] Bulk up test cases with language model suggestions
- [x] Remove `JointEncoder` in favor of just `JointEncoders`
- [x] Consider making it the standard to initialize the value of stateful quantities in `on_configure` instead of `__init__`
- [x] Remove `x_hat` from the attributes of the `ObeliskController`. Let the user manage the internals themselves.
- [x] Change default ROS config string param names to be have suffix `settings` instead of `config_str`.
- [x] Comment out auto sensor list publisher/subscriber creation in `ObeliskRobot` and `ObeliskSensor`

Future
- [ ] Add a `required_fields` argument to all component creation functions to enforce conditionally required fields. This is useful for, e.g., when we want to help the user make a bunch of publishers from a list of config strings.
- [ ] Potential TODO: Add checks that restrict the message types of publishers/subscribers depending on the `ObeliskNode` type (for instance, in the `ObeliskController`, we should restrict that the publisher can only publisher `ObeliskControlMsg` types)


Zolkin1:
Is there a reason your ObeliskRobot doesn't implement an on_shutdown?

obelisk/python/obelisk_py/node.py
```diff
@@ -1,86 +1,515 @@
...
+            )
+            msg_type = check_and_get_obelisk_msg_type(self, config_dict["msg_type"], ObeliskAllowedMsg)
+
+        return msg_type
+
+    def _get_callback_group_from_config_dict(self, config_dict: Dict) -> Optional[CallbackGroup]:
+        """Get the callback group from a configuration dictionary."""
+        if "callback_group" in config_dict:
+            assert isinstance(config_dict["callback_group"], str), "The 'callback_group' field must be a string!"
+            if config_dict["callback_group"].lower() == "none":
```
Zolkin1, line 178: Might want to make "none" -> "None" because iirc all the other fields are CamelCase.

obelisk/python/obelisk_py/node.py
```diff
@@ -1,86 +1,515 @@
...
+        """
+        # parse and check the configuration string
+        field_names, value_names = ObeliskNode._parse_config_str(config_str)
+        required_field_names = ["topic"]
+        optional_field_names = ["msg_type", "history_depth", "callback_group", "non_obelisk"]
+        ObeliskNode._check_fields(field_names, required_field_names, optional_field_names)
+        config_dict = dict(zip(field_names, value_names))
+
+        # parse the message type
+        if msg_type is None:
```
Zolkin1, line 255: So there is a message type argument in this function, but if its not provided, you grab it from the config string? Just want to make sure I understand it correctly

obelisk/python/obelisk_py/robot.py
```diff
@@ -0,0 +1,207 @@
...
+        """Clean up the robot."""
+        super().on_cleanup(state)
+
+        # destroy publishers + config strings
+        self.subscriber_ctrl.destroy()
+        for sensor_publisher in self.publisher_sensors:
+            self.destroy_publisher(sensor_publisher)
+
+        del self.subscriber_ctrl
+        del self.publisher_sensors
```
Zolkin1, line 57: in this version of the code, is it ok for you to manage this delete, or should the user do it since they make the objects?

Zolkin1 (commented):
lgtm. left a few minor comments. only skimmed the tests.

obelisk/python/obelisk_py/node.py
alberthli: this is fine, the `.lower()` call forcibly lower cases everything to check string matching. I think it's actually ok if the user passes things like `none` as well.

obelisk/python/obelisk_py/node.py
alberthli: Yes, that's correct

obelisk/python/obelisk_py/robot.py
alberthli: I think it's fine for us to delete these, since they're the defaults. Down the line, we should expect the user to manage their own deletions. In fact, it's probably OK if they don't even do this, it's pretty benign.

PR #27, opened 2024-06-26
Obelisk C++ library

Zolkin1:
Implementing the C++ side of the Obelisk generic ROS2 interfaces.

- [x] ObeliskNode
- [x] ObeliskController
- [x] ObeliskEstimator
- [x] ObeliskSensor
- [x] ObeliskRobot
- [x] ObeliskSimRobot
- [x] MujocoInterface
- [x] Shielding

TODOs:
- [x] Update all "settings" to be "setting" when referring to a single array
- [x] Add logging for each node state change (include node's name) 
- [x] Make all the on_* functions virtual so they can be overriden
- [x] Fix timing conversion issues
- [x] Deal with publisher activation warnings (there might be a better way to do this...)
- [x] Deal with graceful exit on the mujoco sim
- [x] xhat not being populated
- [x] Change all `REQUIRE` in Catch tests to `CHECK` so that we don't get cascading issues.
- [x] Clear up warning from the GLFW callbacks when using ros-build


obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,451 @@
...
+  public:
+    explicit ObeliskNode(
+        const std::string& node_name,
+        const rclcpp::NodeOptions& options  = rclcpp::NodeOptions(),
+        bool enable_communication_interface = true)
+        : LifecycleNode(node_name, options, enable_communication_interface),
+          CB_GROUP_NONE("None"),
+          CB_GROUP_MUTUALLY_EXEC("MutuallyExclusiveCallbackGroup"),
+          CB_GROUP_REENTRANT("ReentrantCallbackGroup") {
+        this->declare_parameter<std::string>("callback_group_config_strs", "");
```
alberthli, line 32: another reminder to declare the type explicitly here instead of supplying a bad default value. i'll stop commenting on this.

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,451 @@
...
+        typename AllocatorT    = std::allocator<void>,
+        typename SubscriptionT = rclcpp::Subscription<MessageT, AllocatorT>>
+    std::shared_ptr<SubscriptionT>
+    CreateSubscriptionFromConfigStr(const std::string& config,
+                                    CallbackT&& callback) {
+        // Parse the configuration string
+        const auto config_map                = ParseConfigStr(config);
+        const std::string topic              = GetTopic(config_map);
+        const int depth                      = GetHistoryDepth(config_map);
+        const bool non_obelisk               = !GetIsObeliskMsg(config_map);
```
alberthli, line 150: why are these so far to the right?

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,451 @@
...
+
+        rclcpp::CallbackGroup::SharedPtr cbg = nullptr; // default group
+        try {
+            // Get the callback group based on the string name
+            cbg = callback_groups_.at(GetCallbackGroupName(config_map));
+        } catch (const std::exception& e) {
+        }
```
alberthli, line 157: is it chill style-wise to catch e and do nothing? at least in python they complain about this

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,451 @@
...
+
+    /**
+     * @brief Create a wall timer from a configuration string
+     *
+     * @param config the configuration string
+     * @param callback the callback function
+     *
+     * @return the timer
+     */
+    template <typename DurationT = std::milli, typename CallbackT>
```
alberthli, line 193: does this mean you take milliseconds by default? I thought you were doing something finer (i think we should do at least micro)

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,451 @@
...
+                                   obelisk_estimator_msgs::msg::EstimatedState,
+                                   obelisk_sensor_msgs::msg::JointEncoder,
+                                   obelisk_sensor_msgs::msg::JointEncoders,
+                                   obelisk_sensor_msgs::msg::TrueSimState>;
+    // Allowed non-obelisk message types
+    using ROSAllowedMsgs = std::tuple<rcl_interfaces::msg::ParameterEvent>;
+
+    /**
+     * Helper functions for determining if we have a valid message.
+     * These functions recursively dissamble a tuple to check if the type
```
alberthli, line 433: sp: disassemble

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,451 @@
...
+     * matches. See this stack overflow post:
```
alberthli, line 434: sp: stackoverflow

alberthli (commented):
Left some comments. Control code looks mostly good to me in terms of functionality. Didn't look at the tests.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,132 @@
...
+namespace obelisk {
+
+template <typename ControlMessageT, typename EstimatorMessageT>
+class ObeliskController : public ObeliskNode {
+  public:
+    explicit ObeliskController(const std::string& name) : ObeliskNode(name) {
+        // Declare all paramters
+        this->declare_parameter<std::string>("timer_ctrl_config_str", "");
+        this->declare_parameter<std::string>("pub_ctrl_config_str", "");
+        this->declare_parameter<std::string>("sub_est_config_str", "");
```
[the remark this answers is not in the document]
Zolkin1: Yea so actually I couldn't seem to get away without supplying a default. Like if I didn't supply a default, then even if I set them before using them I got a runtime error that I couldn't get around. So I settled to leaving a default empty string. If they are not set then my code will throw an error as all timer configs require a period, and pubs and subs require a topic. So the end result of throwing an error if untouched still happens.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,132 @@
...
+        control_publisher_ = CreatePublisherFromConfigStr<ControlMessageT>(
+            this->get_parameter("pub_ctrl_config_str").as_string());
+
+        state_estimator_subscriber_ =
+            CreateSubscriptionFromConfigStr<EstimatorMessageT>(
+                this->get_parameter("sub_est_config_str").as_string(),
+                std::bind(&ObeliskController::UpdateXHat, this,
+                          std::placeholders::_1));
+
+        control_timer_ = CreateWallTimerFromConfigStr(
```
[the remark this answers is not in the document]
Zolkin1: Ok yea so I could never get the "default" timer to work. Only the wall timer would ever work and thats what is used in all the ros examples in their docs, so I just kinda assumed that maybe this is the one we are supposed to use. That being said I can go back and take another look.

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
Zolkin1: Modified the code to not have empty catch statements.

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
Zolkin1: yea so this is actually the ros default which is why I ultimately went with it, but I can change it to micro seconds.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,103 @@
...
+         *
+         * @param prev_state the state of the ros node.
+         * @return success if everything completes.
+         */
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual on_configure(
+            const rclcpp_lifecycle::State& prev_state) {
+            ObeliskNode::on_configure(prev_state);
+
+            return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+        }
```
alberthli, line 34: Do we need to implement this? It's just one super call. On the python side, we don't even have this and calling super just straight up calls the base class' `on_configure`. We can probably resolve this at the same time as #35.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,103 @@
...
+         * @brief activates the node.
+         *
+         * @param prev_state the state of the ros node.
+         */
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual on_activate(
+            const rclcpp_lifecycle::State& prev_state) {
+            this->ObeliskNode::on_activate(prev_state);
+
+            return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+        }
```
alberthli, line 46: same as above comment on `on_configure`. I'll stop commenting on these callbacks now.

obelisk/cpp/obelisk_cpp/include/obelisk_ros_utls.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
```
alberthli, line 1: the name of this file seems misspelled

obelisk/cpp/obelisk_cpp/include/obelisk_ros_utls.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <string>
+
+#include "rclcpp/rclcpp.hpp"
+
+namespace obelisk::utils {
+    /**
+     * @brief spins up an ObeliskNode.
+     *  Handles initilizng rclcpp, creating the executor, adding the node, spinning, removing the node, and shuting
```
alberthli, line 10: sp: initializing

alberthli: sp: shutting

obelisk/cpp/obelisk_cpp/include/obelisk_sensor.h
```diff
@@ -0,0 +1,56 @@
...
+      public:
+        explicit ObeliskSensor(const std::string& name) : ObeliskNode(name) { has_sensor_pub_ = false; }
+
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual on_configure(
+            const rclcpp_lifecycle::State& prev_state) {
+            ObeliskNode::on_configure(prev_state);
+
+            // TODO (@zolkin): Use the MESSAGE_NAME field (stored in info.msg_type in registered_publishers_) to
+            // determine if the publisher is publishing a sensor
+            //  for now, just going to check if its empty
```
alberthli, line 16: nit: remove extra space in this comment

alberthli, line 16: is there some way to do what the python side does here? we're just checking whether at least one of the publisher message types is a sensor message, and if not, we throw a runtime error

[https://github.com/Caltech-AMBER/obelisk/blob/6a4dc6259b04a9d192f6eb90435e0aab88ac85ed/obelisk/python/obelisk_py/core/sensing.py](https://github.com/Caltech-AMBER/obelisk/blob/6a4dc6259b04a9d192f6eb90435e0aab88ac85ed/obelisk/python/obelisk_py/core/sensing.py)

obelisk/cpp/obelisk_cpp/include/obelisk_sim_robot.h
```diff
@@ -0,0 +1,139 @@
...
+            // Register Components
+            this->RegisterTimer("timer_true_sim_state_setting", state_timer_key_,
+                                std::bind(&ObeliskSimRobot::PublishTrueSimState, this));
+            this->template RegisterPublisher<TrueSimState>("pub_true_sim_state_setting", state_pub_key_);
+
+            stop_thread_ = false;
+        }
+
+        /**
+         * @brief Configures all the required ROS components. Specifcially this
```
alberthli, line 24: sp: specifically

we should somehow add a spellchecker

obelisk/cpp/obelisk_cpp/include/obelisk_sim_robot.h
```diff
@@ -0,0 +1,139 @@
...
+
+        // ---------- Member Variables --------- //
+        // Hold the thread where the simulation takes place
+        std::thread sim_thread_;
+
+        // Flag to signal stopping the thread
+        std::atomic<bool> stop_thread_;
+
+        std::string state_timer_key_;
+        std::string state_pub_key_;
```
alberthli, line 135: On the python side i ended up moving this logic to the mujoco-specific sim class, mostly because the multiprocessing stuff felt a bit too specific to always apply to the generic sim robot API. We should discuss where this should go.

obelisk_ws/src/cpp/obelisk_control_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# find dependencies
+find_package(ament_cmake REQUIRED)
+# uncomment the following section in order to fill in
+# further dependencies manually.
+# find_package(<dependency> REQUIRED)
```
alberthli, line 12: remove these comments

obelisk_ws/src/cpp/obelisk_control_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  # the following line skips the linter which checks for copyrights
+  # comment the line when a copyright and license is added to all source files
+  set(ament_cmake_copyright_FOUND TRUE)
+  # the following line skips cpplint (only works in a git repo)
+  # comment the line when this package is in a git repo and when
+  # a copyright and license is added to all source files
+  set(ament_cmake_cpplint_FOUND TRUE)
+  ament_lint_auto_find_test_dependencies()
```
alberthli, line 35: remove boilerplate comments in these testing blocks

obelisk_ws/src/cpp/obelisk_control_cpp/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
alberthli, line 1: do we need a license file here? i think we already have one at the top level

obelisk_ws/src/cpp/obelisk_control_cpp/src/example_position_setpoint_controller.cpp
```diff
@@ -0,0 +1,34 @@
+#include "rclcpp/rclcpp.hpp"
```
alberthli, line 1: can we have a `zoo` structure in the main `obelisk` package for this?

obelisk_ws/src/cpp/obelisk_estimation_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# find dependencies
+find_package(ament_cmake REQUIRED)
+# uncomment the following section in order to fill in
+# further dependencies manually.
+# find_package(<dependency> REQUIRED)
```
alberthli, line 12: some more boilerplate - won't mention it again

obelisk_ws/src/cpp/obelisk_estimation_cpp/src/jointencoders_passthrough_estimator.cpp
```diff
@@ -0,0 +1,36 @@
+#include "rclcpp/rclcpp.hpp"
+
+#include "obelisk_estimator.h"
+#include "obelisk_ros_utls.h"
+
+class JointEncodersPassthrough : public obelisk::ObeliskEstimator<obelisk_estimator_msgs::msg::EstimatedState> {
```
alberthli, line 6: This class name doesn't match the python side. we need to pick and stick with a convention.

pixi.toml
```diff
@@ -24,6 +24,7 @@ scripts = ["obelisk_ws/install/setup.sh"]
 [dependencies]
 python = ">=3.10"
 pip = ">=24.0"
+yaml = ">=0.2.5,<0.3" # TODO: Remove
```
alberthli, line 27: remove this

pixi.toml
```diff
@@ -129,6 +130,12 @@ pyright = ">=1.1.368"
 cmake = "3.16.*"
 ninja = "1.12.1"
 make = ">=4.3,<5"
+glfw = "3.4"    # TODO: move
+#mesa-libgl-cos6-x86_64 = "11.*"
+#libglu = "9.*"
+# TODO: sudo apt-get update
+#   sudo apt-get install mesa-common-dev
```
alberthli, line 137: deal with this before merge

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -0,0 +1,635 @@
...
+         * @param prev_state the state of the ros node.
+         */
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual on_shutdown(
+            const rclcpp_lifecycle::State& prev_state) {
+            this->ObeliskSimRobot<ControlMessageT>::on_shutdown(prev_state);
+            configuration_complete_ = false;
+
+            // Reset data
+            nu_          = -1;
+            num_sensors_ = 0;
```
alberthli, line 122: these are redundant because you call `on_cleanup` already

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -0,0 +1,635 @@
...
+
+        /**
+         * @brief cleans up the node.
+         *
+         * @param prev_state the state of the ros node.
+         */
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual on_cleanup(
+            const rclcpp_lifecycle::State& prev_state) {
+            this->ObeliskSimRobot<ControlMessageT>::on_cleanup(prev_state);
+            configuration_complete_ = false;
```
alberthli, line 101: what's the point of this variable?

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -0,0 +1,635 @@
...
+            glfwSetMouseButtonCallback(window_, ObeliskMujocoRobot::MouseButtonCallback);
+            glfwSetScrollCallback(window_, ObeliskMujocoRobot::ScrollCallback);
+
+            mujoco_setup_ = true;
+            RCLCPP_INFO_STREAM(this->get_logger(), "Starting Mujoco simulation loop.");
+
+            while (!this->stop_thread_) {
+                if (!glfwWindowShouldClose(window_)) {
+                    //  Assuming MuJoCo can simulate faster than real-time, which it usually can,
+                    //  this loop will finish on time for the next frame to be rendered at 60 fps.
```
alberthli, line 194: nit: we dont necessarily have to render at 60fps right

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -0,0 +1,635 @@
...
+            }
+
+            // TODO: Put this here or in cleanup/shutdown?
+            // free visualization storage
+            mjv_freeScene(&scn);
+            mjr_freeContext(&con);
+
+            // free MuJoCo model and data
+            mj_deleteData(data_);
+            mj_deleteModel(model_);
```
alberthli, line 240: cleanup/shutdown. the `stop_thread_` variable should also get set in `on_cleanup` so that this loop stops in its own thread.

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -0,0 +1,635 @@
...
+                if (sensor_type == "jointpos") {
+                    // Make a key
+                    const std::string sensor_key = "sensor_group_" + std::to_string(num_sensors_);
+
+                    // Make a publisher and add it to the list
+                    auto pub = ObeliskNode::create_publisher<obelisk_sensor_msgs::msg::JointEncoders>(topic, depth);
+                    this->publishers_[sensor_key] =
+                        std::make_shared<internal::ObeliskPublisher<obelisk_sensor_msgs::msg::JointEncoders>>(pub);
+
+                    // TODO: Do we want them all in their own reentrant cbg?
```
alberthli, line 436: no, they can all belong to the one reentrant cbg. if they're in different cbgs, it's functionally the same thing (regardless of whether the other cbgs are reentrant or mutually exclusive)

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,820 @@
...
+        };
+
+    } // namespace internal
+
+    /**
+     * ObeliskNode class.
+     * A lifecycle node whose publishers and subscribers can only publish and
+     * subscribe to Obelisk messages.
+     */
+    class ObeliskNode : public rclcpp_lifecycle::LifecycleNode {
```
alberthli, line 99: once other details match up, we should try to also match the docstrings between the cpp and python version

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,820 @@
...
+
+            // Clean up publishers
+            for (auto& [key, pub] : publishers_) {
+                if (pub) {
+                    pub->Release();
+                }
+            }
+            publishers_.clear();
+            registered_publishers_
+                .clear(); // TODO: Consider if this should be called, or if I should leave this in place
```
alberthli, line 406: discuss todo over slack to clarify decision

alberthli (changes requested):
Skimmed, didn't read tests

docker/Dockerfile
```diff
@@ -39,6 +39,12 @@ RUN apt-get update && apt-get install -y \
     rm -rf /var/lib/apt/lists/* && \
     locale-gen en_US.UTF-8
 
+
+# ../scripts/
+# COPY install-mj-source.sh /install-mj-source.sh
+# RUN chmod +x /install-mj-source.sh && /install-mj-source.sh --mujoco-dir /home
```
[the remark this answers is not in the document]
Zolkin1: I actually never got this script to work iirc. So I'll just remove it for now.

docker/docker-compose.yml
```diff
@@ -28,6 +28,7 @@ services:
       - /tmp/.X11-unix:/tmp/.X11-unix
       - $HOME/.Xauthority:$HOME/.Xauthority:rw
       - $HOME/.bashrc:$HOME/.bashrc
+      - $HOME/mujoco:$HOME/mujoco # for now, volume mounting mujoco
```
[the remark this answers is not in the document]
Zolkin1: We really need to resolve #34 to know what to do here. I'll keep the volume mount in until we have a different idea that we want to implement.

docker/install-mj-source.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
Zolkin1: I'm gonna pull it for now, but as noted above, we need to resolve #34.

obelisk/cpp/obelisk_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+# Obelisk Messages
+find_package(obelisk_control_msgs REQUIRED)
+find_package(obelisk_sensor_msgs REQUIRED)
+find_package(obelisk_estimator_msgs REQUIRED)
+
+# ------- Mujoco ------- #
+# Setup Mujoco path
+set(MUJOCO_VERSION "mujoco-3.1.6" CACHE STRING "mujoco version")
+set(MUJOCO_DIR "/home/$ENV{USER}/mujoco/${MUJOCO_VERSION}" CACHE PATH "Path to Mujoco dir")
```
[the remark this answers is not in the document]
Zolkin1: yup, this is tied to the above comments (and thus #34) - I'll do some more thinking about it today. At the end of the day though, there are a number of workable options so this will be somewhat of a preference thing.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,103 @@
+#pragma once
+
+#include "obelisk_node.h"
+
+namespace obelisk {
+
+    template <typename ControlMessageT, typename EstimatorMessageT> class ObeliskController : public ObeliskNode {
+      public:
+        explicit ObeliskController(const std::string& name, const std::string& ctrl_key = "pub_ctrl",
+                                   const std::string& est_key = "sub_est", const std::string& timer_key = "timer_ctrl")
```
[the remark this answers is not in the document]
Zolkin1: Having thought about it some more, since they know type of component they are getting (i.e. by accessing the proper dict, or by using the proper access function), we actually don't need to annotate the type of component in the name of the key. So "sub_est" -> "est" and four you "subscriber_est" -> "est". Let me know what you think about this.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
Zolkin1: Yea I left this in there until we resolve #35 since I don't want to delete it just to re-write it. I think I'll go ahead and implement #35 soon and that will just clear a lot of this up.

obelisk/cpp/obelisk_cpp/include/obelisk_ros_utls.h
Zolkin1: lol good catch

obelisk/cpp/obelisk_cpp/include/obelisk_sim_robot.h
Zolkin1: I'll look into it... spelling isn't my strong suit.

obelisk/cpp/obelisk_cpp/include/obelisk_sim_robot.h
Zolkin1: I moved the shared data to the mujoco specific class, but I assumed that having the simulation in a separate thread, seemed generic enough that every simulation interface would want it. I can't think of sim that you wouldn't want to do this for, so I think we should leave the multi-threading framework here.

obelisk_ws/src/cpp/obelisk_control_cpp/LICENSE
Zolkin1: this came default in the ros package. I just assumed that this was standard practice to have one in every ROS package, but we can remove it if we want.

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
Zolkin1: `configuration_complete_` is just a flag to ensure that the stuff in the sim thread does not attempt to access data before it has been created.

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
Zolkin1: Thinking about this more, I'm gonna keep these here as they are then created and destroyed all in the same place with no other functions touching them. on_cleanup and on_shutdown both end the thread.

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
Zolkin1: I'm gonna leave this here I think. The only reason to remove it is if we want to let the user shutdown or cleanup and then bring back up the node without registering all the components again. But imo, that is the point of the shutting down: that the node is done. So ill leave clearing this here.

obelisk/cpp/obelisk_cpp/include/obelisk_sim_robot.h
Zolkin1: Leaving this here for my own reference: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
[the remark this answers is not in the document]
Zolkin1: Decided to just match the python version for now, but I added a comment on your PR about making local variables to hold these values.

obelisk/cpp/obelisk_cpp/include/obelisk_sensor.h
Zolkin1: So if we want to manage a list of all the allowed types somewhere, then I can do this. I think we want to move away from this in the long run, but for now I can add it in.

obelisk_ws/src/cpp/obelisk_control_cpp/src/example_position_setpoint_controller.cpp
Zolkin1: done

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,101 @@
...
+        /**
+         * @brief configures the node.
+         *
+         * @param prev_state the state of the ros node.
+         * @return success if everything completes.
+         */
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+        on_configure(const rclcpp_lifecycle::State& prev_state) final {
+            ObeliskNode::on_configure(prev_state);
+
```
alberthli, line 30: nit: delete spurious blank lines like this

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,101 @@
...
+            // Register all components
+            this->RegisterTimer("timer_ctrl_setting", timer_key_, std::bind(&ObeliskController::ComputeControl, this));
+            this->RegisterPublisher<ControlMessageT>("pub_ctrl_setting", ctrl_key_);
+            this->RegisterSubscription<EstimatorMessageT>(
+                "sub_est_setting", est_key_, std::bind(&ObeliskController::UpdateXHat, this, std::placeholders::_1));
+        }
+
+        /**
+         * @brief configures the node.
```
alberthli, line 22: open another issue for us to clean the docstrings on both branches. we should sync the docstrings up, capitalize the first letter, use correct punctuation, etc.

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
```diff
@@ -0,0 +1,101 @@
+#pragma once
+
+#include "obelisk_node.h"
+
+namespace obelisk {
+
+    template <typename ControlMessageT, typename EstimatorMessageT> class ObeliskController : public ObeliskNode {
+      public:
+        explicit ObeliskController(const std::string& name, const std::string& ctrl_key = "publisher_ctrl",
```
alberthli, line 9: didn't we decide to change the names of these default keys?

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -0,0 +1,630 @@
...
+#include <cstdarg>
+#include <filesystem>
+
+#include <GLFW/glfw3.h>
+#include <mujoco/mujoco.h>
+
+#include "obelisk_sensor_msgs/msg/joint_encoders.hpp"
+#include "obelisk_sim_robot.h"
+
+// TODO: Fix issue related to ENV XDG_RUNTIME_DIR=/run/user/1000  # Replace 1000 with your user ID
```
alberthli, line 12: we can probably set this in the dev container stuff/setup script

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
alberthli: why was this resolved? the `Reset data` block is still redundant

alberthli (commented):
very minor concerns - once mujoco install stuff is fixed, it's good to merge. we can clean up after

obelisk/cpp/obelisk_cpp/include/obelisk_controller.h
Zolkin1: I actually decided to stick with the python impl default keys, I guess I don't feel too strongly, and these are reasonable.

Zolkin1:
Change the names for the Obelisk Publishers and Subscribers to be more clear.

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
Zolkin1: added

tests/tests_cpp/CMakeLists.txt
```diff
@@ -14,12 +14,23 @@ FetchContent_MakeAvailable(Catch2)
...
-add_executable(dummy_test
-    dummy_test.cpp
-    dummy_test_2.cpp)
-target_link_libraries(dummy_test PRIVATE Catch2::Catch2WithMain)
-target_link_libraries(dummy_test PUBLIC lib1)
+# add_executable(dummy_test
+#     dummy_test.cpp
+#     dummy_test_2.cpp)
+# target_link_libraries(dummy_test PRIVATE Catch2::Catch2WithMain)
+# target_link_libraries(dummy_test PUBLIC lib1)
```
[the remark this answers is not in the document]
alberthli: reminder to delete this random stuff

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -0,0 +1,849 @@
...
+
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual PostCleanup(
+            __attribute__((unused)) const rclcpp_lifecycle::State& prev_state) {
+            return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+        }
+
+        rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn virtual PostShutdown(
+            __attribute__((unused)) const rclcpp_lifecycle::State& prev_state) {
+            return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+        }
```
alberthli, line 446: refer to slack discussion about this - this comment is a placeholder for that pre-merge

alberthli (approved):
LGTM except before merging, we should resolve the discussion around #35 from slack.

Issue #25, opened 2024-06-26
User Shielding


alberthli:
This is a tracker for the user shielding enhancement discussed in person.

Goal: the end user of Obelisk should have to write as little boilerplate code as possible and focus most of their time writing callbacks related to their application. This issue explains a proposal to achieve this.

- [ ] **Auto-configuration of registered component types.** Each component (e.g., publisher, subscriber, and timer) has a large number of settings associated with it. We deal with this by using configuration yaml files, which get parsed down to config strings. Each component should have a config string associated with it, but we want the end user to do the minimal amount of work setting up the component as possible. To handle this, we introduce a concept called _component registration_.

    Registration means that the user specifies (1) the ros parameter name of the components config string, (2) any message types related to the component, (3) any callbacks that must be registered with that component, and (4) a string key identifier for that component. They do this registration in the class constructor. Then, in the `on_configure` method, we realize the value of the associated config string and create the component for them.

    Further, the component will be added to an internal dictionary of components - one for each type. All auto-registered components must be accessed by their key from this dictionary. Example:
    ```
    class MyObeliskNode(ObeliskNode):
        def __init__(self):
            self.register_obk_publisher(
                ros_parameter="my_pub_setting",
                key="my_pub", 
                msg_type=SomePubObeliskMsgType,
            )
            self.register_obk_subscription(
                ros_parameter="my_sub_setting",
                key="my_sub", 
                msg_type=SomeSubObeliskMsgType,
                callback=self.my_sub_callback,
            )
            self.register_obk_timer(
                ros_parameter="my_sub_setting",
                key="my_timer",
                callback=self.my_timer_callback,
            )

        # later on in some function, we can access the components
        self.obk_publishers["my_pub"].publish(some_msg)  # there are already publishers, subscribers, and timers fields in Node
        my_sub = self.obk_subscribers["my_sub"]
        my_timer = self.obk_timers["my_timer"]
    ```
    
- [ ] **Manage initialization and deletion of all internal stateful quantities.** In the same way that we contain all the components in arrays, we should also maintain a dict of all stateful values `internal`. For instance, in our controller, we maintain the most recent estimate of the state as `x_hat`. We can access this from the dict with `internal['key_for_x_hat']`, where `'key_for_x_hat'` is supplied by the user when they write their derived class.

PR #31, opened 2024-07-01
Shielding PR (Python)

### LINKED ISSUE #22: PATH and PYTHONPATH issue
Somehow, on my local filesystem, my `PATH` and `PYTHONPATH` have been prepended with a bunch of information that points the `python` executable to the one associated with the `pixi` dev environment of this project. This breaks using other `conda` environments, and we should figure out where these environment variables are being set and patch it away.

alberthli:
Addresses the python part of #25.

### Changes Besides Main PR:
- Don't store config strings as parameters, which cleans up code in derived classes
- All example derived modules now live in the `zoo` submodule of `obelisk_py`. All core code (e.g., base class definitions, utils, etc.) live in the `core` submodule.
- Update tests to reflect new API + change file structure to match new structure of modules
- Added a `spin_obelisk` utility function that greatly reduces the amount of ROS boilerplate to write when defining an executable. For example, now the entire executable script for a setpoint controller looks like this:
```
from typing import List, Optional

from rclpy.executors import SingleThreadedExecutor

from obelisk_py.core.utils.ros import spin_obelisk
from obelisk_py.zoo.control.example.example_position_setpoint_controller import ExamplePositionSetpointController


def main(args: Optional[List] = None) -> None:
    """Main entrypoint."""
    spin_obelisk(args, ExamplePositionSetpointController, SingleThreadedExecutor)


if __name__ == "__main__":
    main()
```
- Closed #22 with changes to `source_obelisk.sh`.


obelisk/python/obelisk_py/core/node.py
```diff
@@ -71,6 +81,92 @@ def t(self) -> float:
...
 
+    # ###################### #
+    # COMPONENT REGISTRATION #
+    # ###################### #
+
+    def register_obk_publisher(
+        self,
+        ros_parameter: str,
+        key: Optional[str] = None,
+        msg_type: Optional[Type] = None,
```
Zolkin1, line 92: c++ will require the message type here, and will not be able to load it in via the config string. So this is fine to keep in the python, but I just wanted to say that field in the yaml will be unused in the c++ side

Zolkin1 (approved):
lgtm, but I as I commented below there is one stop where you are exposing functionality on the python side that I can't mimic on the C++ side without it being messy. Not sure if we want to merge this rn, or wait until the C++ shielding side is working though.

Also, branch runs fine on my machine at 1khz.

Zolkin1:
Per f2f we need to add a number of warning that tell the user if config string fields are being used or ignored. Specifically, the source code specifications take precedence over the config string values, so if something is specified in the source code, then a warning that tells the user their config value is being ignored should be issued.

Zolkin1:
We are gonna move the mujoco sim out of `zoo` and into `core`. currently thinking under `core/robot/sim/mujoco.py`, but the exact location is still tbd.

obelisk/python/obelisk_py/core/node.py
alberthli: resolved.

obelisk/python/obelisk_py/core/control.py
```diff
@@ -0,0 +1,58 @@
...
+    the control message should be of type ObeliskControlMsg to be compatible with the Obelisk ecosystem.
+    """
+
+    def __init__(self, node_name: str) -> None:
+        """Initialize the Obelisk controller."""
+        super().__init__(node_name)
+        self.register_obk_timer(
+            "timer_ctrl_setting",
+            self.compute_control,
+            key="timer_ctrl",
```
Zolkin1, line 26: For every key that we make, we should also provide a member variable that holds the key. Then, when the user wants to access the component, they can do it through the member variable rather than a hard-coded string. See for example [here](https://github.com/Caltech-AMBER/obelisk/blob/f08a41814d9770f1129363b6e053bb1b3f278962/obelisk/cpp/obelisk_cpp/include/obelisk_controller.h#L97).

Zolkin1:
Added OnConfigure(), etc... methods for the user to implement that are always called at the correct time. We need to add those here to have matching APIs. See #35.

alberthli:
TODO for myself: move the multiprocessing code from the mujoco robot into the generic sim robot class to match the C++ version

obelisk/python/obelisk_py/core/control.py
alberthli: Resolved - but, I did this in a more general way on the python side. The key name is set as a node attribute and an exception is raised if another key with that name exists already (note that this means that even if the component type is different, the key MUST be different!)

See these lines: [link](https://github.com/Caltech-AMBER/obelisk/blob/08fed086bd71b502edecbe83051cb99b346a29ab/obelisk/python/obelisk_py/core/node.py#L404-L405)

obelisk/python/obelisk_py/core/robot.py
```diff
@@ -60,10 +62,31 @@ def __init__(self, node_name: str) -> None:
...
+    def _set_shared_ctrl(self, ctrl: List[float]) -> None:
+        """Set the shared control array.
+
+        This is a convenience function to set the shared control array in a thread-safe manner. This should be called in
+        the apply_control function. The shared control array is used to communicate the control input to the simulator.
+
+        The values can be accessed in the simulator by running
+        ```
+        with self.lock:
+            ctrl = list(self.shared_ctrl)
```
Zolkin1, line 80: Can we change this to `ctrl_key` to match the C++ side and to emphasize that this attribute is a key?

Goes for all the member variables.

Zolkin1 (approved):
One naming related comment.

Zolkin1:
Can you change the `pyproject.toml` description? Right now it says "Repository for fast robust grasp generation"

PR #45, opened 2024-07-11
Update how the C++ libraries are configured with cmake

Zolkin1:
Updating the CMake for the C++ Obelisk library to support being used with `FetchContent`. Now the user can put the following in their CMake to use the library:

```
include(FetchContent)
FetchContent_Declare(
  Obelisk
  SOURCE_DIR $ENV{OBELISK_ROOT}/obelisk/cpp
)

# Lets us Get Obelisk::Core and Obelisk::Zoo
FetchContent_MakeAvailable(Obelisk)
```
Also added namespaced targets and change the names to match the python side.

### DIFF
--- a/obelisk/cpp/CMakeLists.txt
+++ b/obelisk/cpp/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.16)
 
-project(obelisk_cpp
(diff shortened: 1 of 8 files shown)


alberthli (approved):
lgtm

PR #46, opened 2024-07-11
Specify full ROS package name in configuration yaml file

alberthli:
This PR makes it so that the user can specify the full name of the package in the config.yaml (versus specifying the `impl` field)


Zolkin1 (approved):
lgtm

PR #49, opened 2024-07-12
Logging

alberthli:
This PR adds logging capabilities mainly relying on ROS2's native logger.

Note: logging on the python side is a bit broken for now. See this issue we opened: https://github.com/ros2/rclpy/issues/1315.

### DIFF
--- a/docs/source/development.md
+++ b/docs/source/development.md
@@ -89,5 +89,6 @@ The C++ libraries in `obelisk/cpp` are built with CMake.
 
 All `obelisk` C++ libraries are placed in `obelisk/`. Each library should have its own folder. Within that folder there should be an `include` folder, a `CMakeLists.txt`, and the source files (i.e., not the header files - those go in `include`).
 
+(testing)=
 ### Testing
 Unit tests are managed with [`Catch2`](https://github.com/catchorg/Catch2). Ultimately we plan to run the tests with [`CTest`](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html), and therefore all unit tests need to be registered with `CTest`, see [here](https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md). The tests are all placed within `tests/tests_cpp`.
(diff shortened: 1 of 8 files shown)


obelisk/python/obelisk_py/core/utils/launch_utils.py
```diff
@@ -317,3 +319,30 @@ def _single_component_launch_actions(node_settings: Dict, suffix: Optional[int]
...
+def setup_logging_dir(config_name: str) -> str:
+    """
+    Configures the logging directory.
+
+    Checks if there is already an obk_logs directory, and if not, create on.
+    Then creates a folder inside of that will the date and the name of the config.
+    Sets the ROS_LOG_DIR environment variable to this location.
+    """
+    # Check for the directory
+    general_log_file_path = os.getcwd() + "/obk_logs"
```
alberthli, line 333: need to be careful - `getcwd()` will point to the directory where you launch, which might not be in the repo root. if you want the logs to be located somewhere specific, we need to set the `ROS_HOME` variable

Zolkin1: What would you suggest setting `ROS_HOME` to? Seems like you run into the same issue which is that we don't know the path to the user's workspace. So for now I'm just assuming that is where they launched from.

docs/source/development.md
```diff
@@ -89,5 +89,6 @@ The C++ libraries in `obelisk/cpp` are built with CMake.
 
 All `obelisk` C++ libraries are placed in `obelisk/`. Each library should have its own folder. Within that folder there should be an `include` folder, a `CMakeLists.txt`, and the source files (i.e., not the header files - those go in `include`).
 
+(testing)=
```
alberthli, line 92: i removed this before, what does this do?

Zolkin1: This lets us have an internal page reference. So up above I reference this section, which can only happen with this tag.

Zolkin1 (approved):
lgtm @alberthli do you want to review before I merge in? I don't think we are gonna figure out that python logging bug and this is better than nothing.

PR #50, opened 2024-07-12
Keeping the Mujoco time in better sync with real time.

Zolkin1:
Added a thread sleep to keep the sim in better sync with reality


alberthli (approved):
lgtm

PR #60, opened 2024-07-16
Use Cyclone DDS

### LINKED ISSUE #30: Lag when running python dummy loop on vulcan
When running the full launch file for the dummy example on vulcan, both the viz and the sim sensor publisher are very laggy and slow. however, this doesn't happen when I run the same loop locally on a laptop. Need to debug.

alberthli:
Closes #30.


Zolkin1 (approved):
lgtm. worked on my machine

PR #61, opened 2024-07-16
Multinode settings

alberthli:
[WARNING] don't merge this before #60.

Allows multinode configuration. New breaking change: in yaml files, the node settings under ALL node type headings (i.e., `control`, `estimation`, `robot`, and `sensing`) must be in yaml lists, even if there is only a single node!


Zolkin1 (approved):
lgtm

PR #64, opened 2024-07-17
Updating nodes to allow users to pass file paths

Zolkin1:
Updated the python and c++ nodes to have a ROS parameter called `params_path` where the user can file path.


obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -180,6 +180,9 @@ namespace obelisk {
               CB_GROUP_MUTUALLY_EXEC("MutuallyExclusiveCallbackGroup"), CB_GROUP_REENTRANT("ReentrantCallbackGroup") {
             this->declare_parameter<std::string>("callback_group_setting", "");
 
+            // ROS parameter deisgned to let the user feed a file path for their own code
```
alberthli, line 183: sp: designed

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -189,6 +192,11 @@ namespace obelisk {
             : LifecycleNode(node_name, namespace_, options, enable_communication_interface), CB_GROUP_NONE("None"),
               CB_GROUP_MUTUALLY_EXEC("MutuallyExclusiveCallbackGroup"), CB_GROUP_REENTRANT("ReentrantCallbackGroup") {
             this->declare_parameter<std::string>("callback_group_setting", "");
+
+            // ROS parameter deisgned to let the user feed a file path for their own code
```
alberthli, line 196: same sp error

obelisk/python/obelisk_py/core/node.py
```diff
@@ -60,6 +60,8 @@ def __init__(self, node_name: str) -> None:
         """Initialize the Obelisk node."""
         super().__init__(node_name)
         self.declare_parameter("callback_group_settings", "")
+        # ROS parameter deisgned to let the user feed a file path for their own code
```
alberthli, line 63: sp

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
```diff
@@ -3,6 +3,7 @@ onboard:
   control:
     - pkg: obelisk_control_cpp
       executable: example_position_setpoint_controller
+      params_path: /path/to/my/file
```
alberthli, line 6: would be better to actually update the dummy example to use this in some example way. for example, just specifying a param file with a single param and having that somehow be used

alberthli (approved):
will approve, but haven't run any code myself.

PR #66, opened 2024-07-17
Leap hand cpp

gavin-hyl:
Draft PR, no documentation yet.


obelisk/cpp/zoo/robot/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 3.5)
```
alberthli, line 1: this is a pretty low minimum version. let's stick to 3.16 - we're working to standardize this across our codebase.

obelisk/cpp/zoo/robot/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+project(dynamixel_interface LANGUAGES CXX)
+
+message(STATUS "Configuring Dynamixel Interface")
+
+# Find the Dynamixel SDK library
+find_package(ament_cmake REQUIRED)
+find_package(dynamixel_sdk REQUIRED)
+
+include_directories("$ENV{OBELISK_ROOT}/.pixi/envs/dev/include/dynamixel_sdk")
+include_directories("$ENV{OBELISK_ROOT}/.pixi/envs/dev-no-gpu/include/dynamixel_sdk")
```
alberthli, line 11: i feel like this isn't the best way to do this include. the end user may not even be using pixi, so this would fail for them. is there a better way to specify this dep?

alberthli: also, just for future reference, if we do end up doing hacks like this in the future, we shouldn't hardcode env names in these paths, since those could potentially change and break things. There's a `PIXI_ENVIRONMENT_NAME` environment variable after pixi shells are activated that you can instead reference.

obelisk/cpp/zoo/robot/leap_hand_interface.h
```diff
@@ -0,0 +1,126 @@
...
+            if (port_handler_->openPort()) {
+                std::cout << "Opened port\n";
+            } else {
+                std::cerr << "Failed to open port\n";
+            }
+            if (port_handler_->setBaudRate(BAUDRATE)) {
+                std::cout << "Set baud rate to" << BAUDRATE << "\n";
+            } else {
+                std::cerr << "Failed to set baud rate\n";
+            }
```
alberthli, line 45: can we add some whitespace separating these if blocks? the readability isn't very good

obelisk/cpp/zoo/robot/leap_hand_interface.h
```diff
@@ -0,0 +1,126 @@
...
+            for (int i = 0; i < N_MOTORS; i++) {
+                int err = packet_handler_->write1ByteTxRx(port_handler_, i, 64, 1); // Enable torque
+                err |= packet_handler_->write2ByteTxRx(port_handler_, i, 84, 500);  // kP
+                err |= packet_handler_->write2ByteTxRx(port_handler_, i, 82, 10);   // kI
+                err |= packet_handler_->write2ByteTxRx(port_handler_, i, 80, 50);   // kD
```
alberthli, line 50: can we set these numbers to variables instead, so the semantic meaning is clear?

For example, `torque_enable = 1;` or `kP = 500;` would go a long way in terms of readability

alberthli: also, i'm assuming the numbers like 64, 84, etc. are port numbers. Can you leave a comment demonstrating how you know those are the correct numbers for future reference? Otherwise, it's too much of a magic number.

obelisk_ws/src/cpp/obelisk_leap_control_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+cmake_minimum_required(VERSION 3.8)
```
alberthli, line 1: 3.16

obelisk_ws/src/cpp/obelisk_leap_control_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+find_package(ament_cmake REQUIRED)
+
+# Get the Obelisk library
+include(FetchContent)
+FetchContent_Declare(
+  Obelisk
+  SOURCE_DIR $ENV{OBELISK_ROOT}/obelisk/cpp
+)
+
+# Lets us Get Obelisk::Core and Obelisk::Zoo
```
alberthli, line 18: nit: don't capitalize get

obelisk_ws/src/cpp/obelisk_leap_control_cpp/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
alberthli, line 1: I know we've been pretty inconsistent about this throughout the repo, but I'm in favor for now of removing these license declarations in individual packages.

Also, right now, I chose the MIT license, so this license isn't compatible with the outermost declaration.

obelisk_ws/src/cpp/obelisk_leap_control_cpp/src/obelisk_leap_controller.cpp
```diff
@@ -0,0 +1,7 @@
+#include "leap_position_setpoint_controller.h"
+#include "obelisk_ros_utils.h"
+
+int main(int argc, char* argv[]) {
+    obelisk::utils::SpinObelisk<obelisk::LeapPositionSetpointController, rclcpp::executors::MultiThreadedExecutor>(
```
alberthli, line 5: use single-threaded executors unless we really need the multithreading

obelisk_ws/src/cpp/obelisk_leap_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+cmake_minimum_required(VERSION 3.8)
```
alberthli, line 1: 3.16

obelisk_ws/src/cpp/obelisk_leap_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+FetchContent_Declare(
+  Obelisk
+  SOURCE_DIR $ENV{OBELISK_ROOT}/obelisk/cpp
+)
+
+# Lets us Get Obelisk::Core and Obelisk::Zoo
+FetchContent_MakeAvailable(Obelisk)
+
+include_directories("$ENV{OBELISK_ROOT}/.pixi/envs/dev/include/dynamixel_sdk")
+include_directories("$ENV{OBELISK_ROOT}/.pixi/envs/dev-no-gpu/include/dynamixel_sdk")
```
alberthli, line 24: same comment as above. avoid using pixi include dirs here in case end user isn't using pixi to manage deps. we need some other way to specify this.

obelisk_ws/src/cpp/obelisk_leap_cpp/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
alberthli, line 1: same comment about license.

obelisk_ws/src/cpp/obelisk_leap_cpp/src/obelisk_leap_robot.cpp
```diff
@@ -0,0 +1,10 @@
+#include "rclcpp/rclcpp.hpp"
+
+#include "leap_hand_interface.h"
+#include "obelisk_ros_utils.h"
+#include "rclcpp_lifecycle/lifecycle_node.hpp"
+
+int main(int argc, char* argv[]) {
+    obelisk::utils::SpinObelisk<obelisk::ObeliskLeapHand, rclcpp::executors::MultiThreadedExecutor>(argc, argv,
```
alberthli, line 8: same comment about single vs. multi-threaded executors.

pixi.toml
```diff
@@ -94,6 +94,9 @@ ros-humble-foxglove-bridge="*"
 # ros-humble-rqt = "*"
 # ros-humble-test-msgs = "*"
 
+ros-humble-dynamixel-sdk = "*"
```
alberthli, line 97: any particular reason this isn't version-floored?

obelisk_ws/src/obelisk_ros/config/leap_cpp.yaml
```diff
@@ -0,0 +1,113 @@
...
+            - rf_pip_sensor
+            - rf_dip_sensor
+            - th_cmc_sensor
+            - th_axl_sensor
+            - th_mcp_sensor
+            - th_ipl_sensor
+      timers:
+        - ros_parameter: timer_sensor_setting
+          timer_period_sec: 0.02
+          callback_group: None
```
alberthli, line 113: delete this unnecessary timer. Also delete it on `leap_py.yaml`, I just noticed it.

alberthli (changes requested):

pixi.toml
gavin-hyl: I tried to floor it to the current version but that caused an error, I'll try to use a lower version.

obelisk_ws/src/obelisk_ros/config/leap_cpp.yaml
gavin-hyl: I need a timer to read and publish the state of the robot.

obelisk/cpp/zoo/robot/CMakeLists.txt
gavin-hyl: Yes, using `<package>_INCLUDE_DIRS` solved the issue.

pixi.toml
alberthli: resolved in slack.

obelisk_ws/src/obelisk_ros/config/leap_cpp.yaml
alberthli: resolved in slack.

pixi.toml
```diff
@@ -121,7 +124,7 @@ pytest-xdist = ">=3.6.1"
 typeguard = ">=4.3.0"
 
 # Dynamixel SDK for LEAP hand
-dynamixel-sdk = ">=3.7"
+dynamixel-sdk = ">=3.7.0"
```
alberthli, line 127: i would actually move the dynamixel dependency out of this dependency group (which is meant for testing). you can create a new dependency group called "leap" (do this by defining a new toml header `[feature.leap.pypi-dependencies]`) and then putting `dynamixel-sdk` under it. You should then make sure `dev` and `dev-no-gpu` include the `leap` dependency group.

obelisk_ws/src/cpp/obelisk_leap_control_cpp/CMakeLists.txt
alberthli: this wasn't resolved: I'm talking about the "Get" in this comment:

> Lets us Get Obelisk::Core and Obelisk::Zoo

obelisk_ws/src/cpp/obelisk_leap_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+include(FetchContent)
+FetchContent_Declare(
+  Obelisk
+  SOURCE_DIR $ENV{OBELISK_ROOT}/obelisk/cpp
+)
+
+# Lets us get Obelisk::Core and Obelisk::Zoo
+FetchContent_MakeAvailable(Obelisk)
+
+set(LD_LIBRARY_PATH "/home/gavin/caltech/surf2024/obelisk/.pixi/envs/dev/lib:$ENV{LD_LIBRARY_PATH}")
```
alberthli, line 23: this LD_LIBRARY_PATH is hardcoded to your local filesystem path - you can remove this line, since we patched the fix on the pixi side for needing to edit the LD_LIBRARY_PATH already.

obelisk_ws/src/cpp/obelisk_leap_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+include_directories("${dynamixel_sdk_INCLUDE_DIRS}/dynamixel_sdk")
+
+# export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/gavin/caltech/surf2024/obelisk/.pixi/envs/dev/lib/
```
alberthli, line 26: remove this comment

pixi.toml
```diff
@@ -7,6 +7,7 @@ name = "obelisk"
 authors = [
     "Albert Li <alberthli@caltech.edu>",
     "Zach Olkin <zolkin@caltech.edu>",
+    "Gavin Hua <ghua@caltech.edu>",
```
alberthli, line 10: Nice!

alberthli (changes requested):
Looks mostly good, some minor things left.

Main thing is that you should make sure this works both in pixi and with local system installation (which you can safely test in docker). You should also verify the sim AND hardware behavior on both pixi and local system, too.

obelisk/cpp/zoo/robot/CMakeLists.txt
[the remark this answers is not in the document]
gavin-hyl: When I run `obk-clean` after installing deps my shell can't find the command - just pulled main, is there anything else I need to set up?

gavin-hyl:
Tested the leap hand behavior in all combinations of sim/hardware and pixi shell/no shell and everything is as expected.

alberthli (approved):

PR #68, opened 2024-07-19
New LEAP Hand Model

alberthli:
This PR updates the LEAP hand model.


gavin-hyl (approved):
XML makes sense, both the C++ and the python code work with this new model.

PR #69, opened 2024-07-20
Fix GL Error

alberthli:
This PR resolves two outstanding issues at their root:

- Seemingly not being able to find rosidl-related packages
- Missing GL dependencies during build time

The main issue with the first problem was that we were mixing system-level compilation deps with pixi compilation deps. If we remove the system-level deps and let conda handle it, we're OK. This also means we don't need the workaround of editing the `LD_LIBRARY_PATH`.

For the second issue, simply adding `mesalib` as a dependency in pixi was enough.


Zolkin1 (approved):
lgtm

PR #74, opened 2024-07-21
Trim Some Dependencies

### LINKED ISSUE #73: Trim System-Level Deps and Minimize System/`~/.bashrc` Pollution with Setup Scripts
We should seek to minimize the impact that setting up Obelisk has on the end user's local filesystem.

- We should minimize the system-level development dependencies as much as possible by moving deps into pixi when we can.
    - However, we should make sure that if the end user doesn't want to use pixi, there is a clear way for them to do so.
    - Some notes on this front
        - `mesa-common-dev` seems hard to eliminate. We can build the `glfw`-related packages in `pixi` by specifying `glfw` and `mesalib` as `conda` dependencies, but we ultimately cannot open the `glfw` window without locally installing `mesa-common-dev`. 
        - We might be able to eliminate `libyaml-dev` as a system-level dependency, but we haven't investigated this thoroughly on the `conda` side.
- The setup script and pixi tasks currently add a lot to the user's bashrc. We may want to consider adding all these things to a separate file that the end user can choose to source temporarily but is sourced automatically by pixi to reduce the impact we have on the end user's settings.

alberthli:
This PR closes #73.

- Removes `uv` and `nvm` as system-level dependencies installed by `docker_setup.sh`.
- Moves `PYRIGHT_PYTHON_FORCE_VERSION`, `RCUTILS_COLORIZED_OUTPUT`, and `RMW_IMPLEMENTATION` environment variables into the `pixi.toml` to prevent `bashrc` pollution.
- Moves obelisk aliases out of the `~/.bashrc` and into the `~/.bash_aliases` file to reduce pollution
- Removes the `source-obelisk` task, which didn't do what we wanted to. Now, introduces a new `obk-build` alias that properly builds and sources the obelisk packages in the correct order. Works both in and out of a pixi shell!
- Removes `libyaml-dev` as a dep
- Mount `~/.gitconfig` and `~/.ssh` into Docker containers so end user can develop properly (outside the .devcontainer context)
- Add a Docker compose file specifically for CI (doesn't do ssh stuff)
- Adds the ability to do pixi-free building of the obelisk packages via the `install_sys_deps.sh` script
- Modifies the setup scripts such that any local changes made to system are always opt-in
- Update README with information on how to set up, build, and use Obelisk
- Added `obk-clean` alias that cleans out the build/install/log dirs of obelisk_ws
- remove cuda 11.8 dep set from pixi
- **Sets up `obk` command that sets environment variables in the current shell. Also builds Obelisk if it's not already built. MOST IMPORTANT NEW ADDITION!**
    - source base ros if not in a conda or a pixi environment and `/opt/ros/humble/setup.bash` exists when running `obk`

Unanswered questions:
- Are there issues with building obelisk via pixi and then sourcing it in downstream projects? Need to evaluate in a future example.

TODOs:
- update docs


Zolkin1 (approved):
Looks good to me. All my questions were answered in code. Verified main repo, example repo, and person downstream repo.

PR #80, opened 2024-07-23
Configure Additional Declared ROS Parameters in YAML File

alberthli:
Sometimes, we may want to expose additional ROS parameters in an ObeliskNode. It may be convenient to set these directly in the master yaml file rather than supplying a path to some additional configuration file, especially if there are only a few parameters. This PR allows the end user to configure these parameters.

### DIFF
--- a/obelisk/cpp/zoo/include/position_setpoint_controller.h
+++ b/obelisk/cpp/zoo/include/position_setpoint_controller.h
@@ -13,6 +13,7 @@ class PositionSetpointController : public obelisk::ObeliskController<obelisk_con
     PositionSetpointController(const std::string& name)
         : obelisk::ObeliskController<obelisk_control_msgs::msg::PositionSetpoint,
                                      obelisk_estimator_msgs::msg::EstimatedState>(name) {
+        // example of params_path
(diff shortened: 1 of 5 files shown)


obelisk/cpp/zoo/include/position_setpoint_controller.h
```diff
@@ -28,6 +29,12 @@ class PositionSetpointController : public obelisk::ObeliskController<obelisk_con
         std::string contents;
         params_file >> contents;
         amplitude_ = std::stof(contents);
+
+        // example of setting other configurable params in the node
+        this->declare_parameter<std::string>("test_param", "default_value");
+        std::string test_param_value_;
```
Zolkin1, line 35: nit: we only use the "_" at the end for member variables. but for local variables within a member function you can just use normal snake case.

Zolkin1 (approved):
lgtm. verified all three examples locally and all worked.

obelisk/cpp/zoo/include/position_setpoint_controller.h
alberthli: resolved

PR #81, opened 2024-07-23
Mujoco cpp sensors

Zolkin1:
Added the following messages:

- ObkImu
- ObkFramePose
- ObkPDFeedForward (not used anywhere for now)

Renamed: JointEncoders -> ObkJointEncoders. Removed `y` field. Added `joint_names` field, `joint_pos`, `joint_vel` fields.

Added support for these messages in the Mujoco C++ simulation.

Updated examples.

- Dummy bot now has an IMU and FramePose sensor

Added support for sensor of heterogeneous Mujoco sensor types.
Including:

- In the config files mujoco sensor names have mujoco sensor types associated with them
- Mujoco simulation now packs message properly rather than dumping into `y`

Added velocities to EstimatedState


Zolkin1:
@alberthli I believe that the changes to the launch file parsing break the python mujoco side (you would just need to add the extra layer of parsing), its not too bad. I didn't do it because I wasn't sure if we wanted to keep the python side up to date. Other than that, all the python stuff should be up to date.

obelisk_ws/src/obelisk_msgs/obelisk_sensor_msgs/msg/ObkImu.msg
```diff
@@ -0,0 +1,16 @@
+# Obelisk compliant message type to hold IMU information
+
+string MESSAGE_NAME="ObkImu"
+
+std_msgs/Header header
+
+geometry_msgs/Quaternion orientation
+float64[9] orientation_covariance # Row major about x, y, z axes
```
alberthli, line 8: I don't feel like this "makes sense" in the sense of differential geometry.

I think if you're going to define the orientation noise using something like a Gaussian, it should be over something like the elements of the Lie algebra se(3), and noise elements should have the exponential map applied to them before being "added" (in the sense of composing two rotations) back to the original rotation.

The main downside to this is that it makes the definition of the noise (in the tangent space) a bit less intuitive to grok, but the main upside is that it's a bit more mathematically coherent. Also, if the noise is small, elements of the tangent space are very similar to the elements of the actual group (in the ambient space).

obelisk_ws/src/obelisk_msgs/obelisk_sensor_msgs/msg/ObkImu.msg
alberthli, line 1: In this message docstring, you should explain that we can't natively guarantee that the covariances make sense. You should also clarify these are Gaussian noise parameters, and maybe include one usage example to make it super clear what the row-major order should "look like" in terms of the 3x3 matrix.

You could also consider using the new multiarray message types we defined to unambiguously deal with this without the possibility of introducing flattening errors.

obelisk_ws/src/obelisk_msgs/obelisk_sensor_msgs/msg/ObkJointEncoders.msg
```diff
@@ -0,0 +1,7 @@
+string MESSAGE_NAME="ObkJointEncoders"
+
+builtin_interfaces/Time stamp
+
+float64[] joint_pos
+float64[] joint_vel
+string[] joint_names
```
alberthli, line 7: So we require the order and length of `joint_pos` and `joint_vel` to be the same, right? I'm wondering if there's some clean way to enforce that in the message type, but I'm not really seeing it

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
```diff
@@ -82,6 +82,20 @@ onboard:
           sensor_settings:
           - topic: /obelisk/dummy/sensor
             dt: 0.001
-            sensor_type: jointpos
+            sensor_type: ObkJointEncoders
```
alberthli, line 85: if we're going to do this, let's rename the `sensor_type` field to `msg_type` to be unambiguous

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
```diff
@@ -82,6 +82,20 @@ onboard:
...
             sensor_names:
-            - sensor_joint1
+            - joint_pos: jointpos
```
alberthli, line 87: for documentation/tutorial purposes, let's leave a comment here explaining how the dict structure works in this field:

the key is the name of the mujoco sensor as specified in the xml. the value is the field which the mujoco sensor should be associated with in the Obelisk message.

alberthli (commented):
Mostly LGTM - c++ side reviewed + run locally, confirmed things look ok.

TODO: I need to fix up the python side.

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
```diff
@@ -126,16 +129,18 @@ namespace obelisk {
...
         using ObeliskMsgs =
-            std::tuple<obelisk_control_msgs::msg::PositionSetpoint, obelisk_estimator_msgs::msg::EstimatedState,
-                       obelisk_sensor_msgs::msg::JointEncoders, obelisk_sensor_msgs::msg::TrueSimState,
-                       obelisk_sensor_msgs::msg::ObkImage, obelisk_std_msgs::msg::FloatMultiArray,
-                       obelisk_std_msgs::msg::UInt8MultiArray>;
+            std::tuple<obelisk_control_msgs::msg::PositionSetpoint, obelisk_control_msgs::msg::ObkPDFeedForward,
+                       obelisk_estimator_msgs::msg::EstimatedState, obelisk_sensor_msgs::msg::ObkJointEncoders,
+                       obelisk_sensor_msgs::msg::TrueSimState, obelisk_sensor_msgs::msg::ObkImage,
+                       obelisk_sensor_msgs::msg::ObkImu, obelisk_sensor_msgs::msg::ObkFramePose,
+                       obelisk_std_msgs::msg::FloatMultiArray, obelisk_std_msgs::msg::UInt8MultiArray>;
```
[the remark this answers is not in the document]
Zolkin1: yea so I wanted to do that, but the clang pre-commit hook continues to re-do the white space. I can see if I can compose the tuples though.

obelisk_ws/src/obelisk_msgs/obelisk_sensor_msgs/msg/ObkJointEncoders.msg
Zolkin1: also, if you have a joint with a quaternion state, then the velocity joint won't be the same size.

obelisk/cpp/obelisk_cpp/include/obelisk_node.h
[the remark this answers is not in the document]
alberthli: there's probably a way to tell the linter to ignore certain lines by leaving a certain comment in the code (at least this is how python linters work)

obelisk_ws/src/obelisk_msgs/obelisk_sensor_msgs/msg/ObkJointEncoders.msg
alberthli: Yes, agree. We should at the very least document what the expected convention is for these things.

obelisk_ws/src/obelisk_msgs/obelisk_control_msgs/msg/ObkPDFeedForward.msg
```diff
@@ -0,0 +1,5 @@
+string MESSAGE_NAME="ObkPDFeedForward"
```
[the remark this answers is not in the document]
Zolkin1: Ok lets not prefix them

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -410,81 +411,374 @@ namespace obelisk {
...
+                        if (mj_sensor_types.at(i) == "framepos") {
+                            if (!has_framepos) {
+                                msg.position.x = this->data_->sensordata[sensor_addr];
+                                msg.position.y = this->data_->sensordata[sensor_addr + 1];
+                                msg.position.z = this->data_->sensordata[sensor_addr + 2];
+
+                                // Framepos sensors can be mounted to different objects
+                                int obj_type = this->model_->sensor_objtype[sensor_id];
+                                if (obj_type == mjOBJ_SITE) {
+                                    int site_id = this->model_->sensor_objid[sensor_id];
```
alberthli, line 688: the x_id variables here are all defined identically. you can probably define it right under obj_type just once and call it obj_id, passing it as the index in every if block

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -410,81 +411,374 @@ namespace obelisk {
...
+                                            << sensor_names.at(i)
+                                            << ", is not associated with a supported Mujoco object type! "
+                                               "Current object type (mjtObj): "
+                                            << obj_type);
+                                }
+
+                                if (this->model_->sensor_refid[sensor_id] == -1) {
+                                    msg.header.frame_id = "world"; // TODO: Consider not hard-coding this
+                                } else {
+                                    int obj_type = this->model_->sensor_reftype[sensor_id];
```
alberthli, line 716: nit: name this `ref_type` to distinguish it from `obj_type` above

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -410,81 +411,374 @@ namespace obelisk {
...
+                                    if (obj_type == mjOBJ_SITE) {
+                                        int site_id = this->model_->sensor_refid[sensor_id];
```
alberthli, line 718: same comment as above about the ids being clutter

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
```diff
@@ -82,6 +82,21 @@ onboard:
           sensor_settings:
           - topic: /obelisk/dummy/sensor
             dt: 0.001
-            sensor_type: jointpos
+            msg_type: ObkJointEncoders
             sensor_names:
-            - sensor_joint1
+            - joint_pos: jointpos
+              joint_vel: jointvel
```
alberthli, line 89: make the value for these `joint_pos` and `joint_vel` to match the field name in the message type. this means you also need to change the parsing code to look for the underscore version in the mujoco sim robot.

alberthli: also do this for the dummy_cpp_viz robot

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
```diff
@@ -82,6 +82,21 @@ onboard:
           sensor_settings:
           - topic: /obelisk/dummy/sensor
```
alberthli, line 84: rename this topic to be `obelisk/dummy/joint_encoders`

alberthli (approved):
minor comments before merge - my python changes are implemented

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -410,81 +411,374 @@ namespace obelisk {
...
+                            }
+                        } else {
+                            RCLCPP_ERROR_STREAM(
+                                this->get_logger(),
+                                "Sensor " << sensor_names.at(i)
+                                          << " is not associated with a valid Mujoco sensor type! Current sensor type: "
+                                          << mj_sensor_types.at(i));
+                        }
+
+                        GetTimeFromSim(msg.header.stamp.sec, msg.header.stamp.nanosec);
```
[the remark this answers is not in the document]
Zolkin1: So the header file has a frame_id field which gives the relative frame. The JointEncoders don't need that.

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
alberthli: i resolved this

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
alberthli: i resolved this

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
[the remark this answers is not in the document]
alberthli: ok, makes sense to me

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -410,81 +411,374 @@ namespace obelisk {
...
+                return cb;
+            } else if constexpr (std::is_same<MessageT, obelisk_sensor_msgs::msg::ObkImu>::value) {
+                // ----------------------------------- //
+                // ---------- IMU call back ---------- //
+                // ----------------------------------- //
+                auto cb = [publisher, sensor_names, mj_sensor_types, this]() {
+                    obelisk_sensor_msgs::msg::ObkImu msg;
+
+                    // TODO: For now all the covariances are 0
+                    // *** Note *** Mujoco does not support adding noise naitevly since v3.14.
```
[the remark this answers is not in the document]
alberthli: i resolved this

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
alberthli: i resolved this

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
alberthli: i resolved this

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
alberthli: i resolved this.

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -410,81 +411,374 @@ namespace obelisk {
...
+                        } else {
+                            RCLCPP_ERROR_STREAM(
+                                this->get_logger(),
+                                "Sensor " << sensor_names.at(i)
+                                          << " is not associated with a valid Mujoco sensor type! Current sensor type: "
+                                          << mj_sensor_types.at(i));
+                        }
+
+                        GetTimeFromSim(msg.header.stamp.sec, msg.header.stamp.nanosec);
+                    }
```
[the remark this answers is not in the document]
Zolkin1: Yea so right now I enforce that you can't have more sensors, but I think its ok if you only want to group two, like just the accelerometer and gyro.

PR #91, opened 2024-07-24
Hardware ZED Camera Node Part 1 (Installation Overhaul)

### LINKED ISSUE #87: Allow end user to have more control over hardware-specific dependencies
For example, the LEAP hand requires the `dynamixel_sdk`, but you don't need to install this unless you're using the LEAP hand. Similarly, the ZED cameras require installation of the ZED SDK, but non-ZED users shouldn't be forced to install this on their system.

alberthli:
This PR begins the process of implementing a ZED2 Camera Node in Obelisk. We start by overhauling the way dependencies are managed/installed with conditional builds and configuring the packages that we build based on the initial setup flags.

Closes #87.

#### Installation
The major change to installation is that we now expose build args set when running `setup.sh` to the `Dockerfile`, which means the image now conditionally builds system-level dependencies based on environment variables on the system. Example usage:
```
cd $OBELISK_ROOT

# this configures environment variables (including in `docker/.env`) to build the container with certain deps
source setup.sh --recommended --docker-zed

# this now conditionally builds things like the basic system-level deps or zed deps into the container
# based on options passed to setup.sh
cd docker
docker compose -f docker-compose.yml run --build obelisk
```

- modifications to `setup.sh`
    - add a `--zed` and a `--basic` flag. these flags will set `OBELISK_BASIC=true` and `OBELISK_ZED=true`.
    - copy the `install_sys_deps.sh` script into the `docker/` directory
- modifications to `install_sys_deps.sh`
    - expose a `--basic` and `--zed` flag. Basic will install the stuff that's in there now. Zed will install the ZED SDK.
- modifications to `Dockerfile` and `docker-compose` yamls
    - expose the `OBELISK_BASIC` and `OBELISK_ZED` environment variables as build args
    - based on their value, pass `--basic` or `--zed` conditionally into the `install_sys_deps.sh` script, which should be added to the image
- move cyclone performance optimizations to install script
- rename `docker_setup.sh` to `user_setup.sh`, move it to `scripts` directory, and copy it into the `docker` directory upon running `setup.sh`
- Separated out the leap hand deps such that you have to opt into them, i.e., `bash install_system_deps.sh --leap`
- zed python sdk is now installed upon running `obk` if it wasn't installed already and if the sdk files have been downloaded

#### Usage
- removed the `messages-build` and `ros-build` pixi tasks in favor of the `obk-build` and `obk` alias. These tasks are now called `messages-build-ci` and `ros-build-ci`, and are only used for CI
- **[CRITICAL]** Aliases now selectively build ROS packages based on the flags passed to `setup.sh` on the host system (for example, if we specify `--leap`, we install the leap-related packages in obelisk, and if we don't, we exclude those packages). More concrete example:
```
# OPTION 1
# on host
source setup.sh --recommended
# in pixi shell
obk-clean
obk  # 15 packages finished

# OPTION 2
# on host
source setup.sh --recommended --leap
# in pixi shell
obk-clean
obk  # 17 packages finished
```

#### File Structure
- Reorganized LEAP hand python/cpp code to all sit in one unified package (also slightly updated naming for clarity). Future: reorganize example code (like example controllers) in a separate workspace for cleanliness.

#### Functional
- **[Follow-up PR]** Code an `ObeliskSensor` called `Zed2Sensors` which initializes and synchronizes several ZED camera outputs, publishing an `ObkImage` message.

#### Future Issues
- Organize example code in a separate workspace (also in this repo) that the end user can build on top of `obelisk_ws` to reduce package clutter. Issue #94.
- Clearly separate robot-specific packages from generics in the package file structure. Issue #95.
- Certain install paths seem broken as reported by Gavin:
    - `source setup.sh --recommended --zed --leap --install-sys-deps` apparently broken on local system
    - `source setup.sh --recommended --zed --leap --install-sys-deps-docker` fails to install deps in the docker filesystem correctly
- Fix nvidia-container-toolkit install for systems that don't yet even have nvidia drivers (so nvidia-smi can't be run). Issue #96.


README.md
```diff
@@ -10,33 +10,59 @@ Obelisk should be used as a dependency for external robot control code that is w
...
+  --obk-aliases                Add obelisk aliases to the ~/.bash_aliases file
+
+  --help                       Display this help message and exit
 ```
+
+Some guidance/recommendations on choosing flags:
+* If you don't have Docker on your machine, use the `--docker-install` flag
+* If you are not using Docker, then you should use `--install-sys-deps`
+* If you are using Docker, but not pixi, you should also use `--install-sys-deps`
+* If you are using Docker, use `--install-sys-deps-docker` if and only if you are **not** using pixi within the container.
```
Zolkin1, line 49: Reading these two lines is a bit confusing. Can we clarify this a bit. Right now both conditions read the same, but suggest doing different things. I.e. both suggest something different for docker with no pixi.

README.md
```diff
@@ -10,33 +10,59 @@ Obelisk should be used as a dependency for external robot control code that is w
 3. Use Obelisk in a project that uses `pixi`.
 
 ### Initial Setup
```
Zolkin1, line 12: It might be worth clarifying that if the user wants to use Obelisk in a custom docker container (i.e. not the one we provide) then they will want to clone this **inside** the docker container and run the command that modify the "system level deps", because all of the conditional building is really for the dev container and presumably the user won't want to dev their robot stack in the Obelisk dev container. 

On the flip side maybe this is an overreach of the docs, but I think it might make the intended use clearer.

docker/Dockerfile
```diff
@@ -29,22 +41,35 @@ RUN apt-get update && apt-get install -y \
...
-# create non-root user with sudo privileges for certain commands
+# create non-root user with sudo privileges
 RUN groupadd --gid $GID $USER && \
     useradd --uid $UID --gid $GID -m $USER -d /home/${USER} --shell /usr/bin/bash && \
     echo "${USER}:password" | chpasswd && \
     usermod -aG sudo ${USER} && \
     echo "%${USER} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
 
-# add user to dialout group to enable serial port access from the container
-RUN usermod -a -G dialout ${USER}
```
Zolkin1, line 40: Do we really not need this anymore? Do we replace it with anything?

pixi.toml
```diff
@@ -36,12 +36,10 @@ obelisk_py = { path="./obelisk/python", editable=true }
 
 # cuda
 [feature.cuda121]
-channels = ["nvidia", "nvidia/label/cuda-12.1.0"]
-system-requirements = {cuda="12.1.0"}
+system-requirements = {cuda="12.1.1"}
 
 [feature.cuda121.dependencies]
-cuda = { version="*", channel="nvidia/label/cuda-12.1.0" }
-
```
Zolkin1, line 44: What is the idea behind removing these channels?

scripts/build_obelisk.sh
```diff
@@ -1,28 +1,54 @@
...
+
+MESSAGE_PKGS=" obelisk_control_msgs obelisk_estimator_msgs obelisk_sensor_msgs obelisk_std_msgs"
+
+# configuring which packages to skip
+SKIP_PKGS=""
+if [ "$leap" = false ]; then
+    SKIP_PKGS+=" obelisk_leap_cpp obelisk_leap_py"
+fi
+if [ "$zed" = false ]; then
+    SKIP_PKGS+=""  # TODO(ahl): fill this once ZED packages are added
```
Zolkin1, line 34: Pretty sure you've added the ZED packages in so you should be able to fill this out. Still reading the code so I'll see for sure later - just dropping this here so I don't forget.

scripts/build_obelisk.sh
```diff
@@ -1,28 +1,54 @@
...
 fi
 
+# building Obelisk packages
+OBELISK_ROOT=$(dirname $(dirname $(readlink -f ${BASH_SOURCE[0]})))
```
Zolkin1, line 38: Don't we have the OBELISK_ROOT environment variable set by now? Can't you use that?

scripts/build_obelisk.sh
```diff
@@ -1,28 +1,54 @@
...
+
+echo -e "\033[1;32mBuilding Obelisk ROS messages...\033[0m"
+curr_dir=$(pwd)
+cd $OBELISK_ROOT/obelisk_ws
+colcon build --symlink-install --parallel-workers $(nproc) \
+    --packages-select $MESSAGE_PKGS \
+    --packages-skip $SKIP_PKGS
```
Zolkin1, line 45: Seems like we can drop this packages-skip call here since we are only building the messages, so we will never need to skip those.

scripts/build_obelisk.sh
```diff
@@ -1,28 +1,54 @@
...
+
+echo -e "\033[1;32mBuilding remainder of Obelisk ROS packages...\033[0m"
+source $OBELISK_ROOT/obelisk_ws/install/setup.bash
+colcon build --symlink-install --parallel-workers $(nproc) \
+    --packages-skip $MESSAGE_PKGS $SKIP_PKGS
```
Zolkin1, line 50: Might be nice to add a --verbose option to obk-build that will force colcon to dump outputs to the screen (useful for debugging). 

Specifically something like one of the first two options [here](https://colcon.readthedocs.io/en/released/user/how-to.html).

Zolkin1 (commented):
Reviewed all changes, but I haven't run any code, so I'm not approving yet. I'll run it first thing this morning. 

Overall everything looks reasonable, but it would be good to update the downstream example repo with how to use this now. I think that is a pretty good test bed to be sure everything runs and to provide a real world-ish use case example.

docker/Dockerfile
alberthli: this was moved into the leap-specific setup

pixi.toml
alberthli: great question. for some reason, i noticed that having those channels made all the downstream cuda deps much higher versions like 12.5 (this causes problems for zed). if i removed the channels, then suddenly everything installed correctly via pixi. you'll notice that if you do `pixi list | grep cuda` now, all the versions of everything are `12.1.*`, which is what we want

scripts/build_obelisk.sh
alberthli: i haven't yet, those are referring to the ZED ROS2 packages (follow-up PR). I should clarify this comment, though.

scripts/build_obelisk.sh
alberthli: this is just me being redundant. i can imagine some weird case where someone is running the scripts out of order or something for debugging purposes or whatever, and somehow OBELISK_ROOT hasn't been set. it's trivial for me to just set it also in this file.

scripts/build_obelisk.sh
alberthli: you're right, good catch

scripts/build_obelisk.sh
alberthli: It's a good idea, let me just open an issue about it though, since i don't want to brick the colcon build command that is working now

README.md
alberthli: good catch, there was a typo I believe

Zolkin1 (approved):
lgtm

README.md
alberthli: Made a halfway attempt at commenting on this - if we need further clarity, let's resolve in a future PR

PR #92, opened 2024-07-24
Removed line in achilles model

Zolkin1:
Extra line in the Achilles model was causing an error. Thought it was tested before the final push, but apparently this line was missed.

Also fixed the mujoco quaternion parsing code.


alberthli (approved):

alberthli (approved):
Good catch.

PR #103, opened 2024-07-26
Added ability to give Mujoco initial conditions from the keyframes.

Zolkin1:
Added ability to start simulation from Mujoco key frames.

Note that key frame are automatically checked for compatibility by Mujoco so there is no need to verify sizes.

Currently only support `qpos`, `qvel`, `ctrl`, and `time`.


gavin-hyl (approved):
Everything makes sense! One small question, why is the convention in this repo `self.declare_parameter` instead of just `self.param = ...`?

PR #104, opened 2024-07-26
Hardware Zed Cam Sensor

### LINKED ISSUE #33: Building ROS (i.e. pixi run ros-build) prints out many warnings
```
--- stderr: obelisk_ros
/home/zolkin/AmberLab/Project-Obelisk/obelisk/.pixi/envs/dev/lib/python3.11/site-packages/setuptools/command/develop.py:42: EasyInstallDeprecationWarning: easy_install command is deprecated.
!!
        ********************************************************************************
        Please avoid running ``setup.py`` and ``easy_install``.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.
        See https://github.com/pypa/setuptools/issues/917 for details.
        ********************************************************************************
!!
  easy_install.initialize_options(self)
---
```

Is printed for many of the Obelisk ros packages. It floods the screen and makes it hard to see other errors.

### LINKED ISSUE #54: Remove `final` from all the on_* methods in the C++ implementation
This is a left over from when we had the Post* functions. We need to remove the `final` keyword so they can be overridden by the user.

alberthli:
This PR finally implements the ObeliskSensor node for ZED2 cameras.

### Changes

- closes #54 
- closes #33 
- updated `msg_conversions.h` to support row-major multiarray messages while sticking to the default column-major Eigen tensor convention
- fixed issues with improper configures/builds due to missing `add_subdirectory` calls in `CMakeLists.txt` files
- changed dynamixel control target to be a library instead of an executable
- C++ multi-camera ZED node working, publishing at proper rates confirmed
    - In particular, we can publish `ObkImage` messages without using `image_transport` at high rates
    - Confirmed that it's easy to subscribe to and view the images. For example:
        ```
        import matplotlib.pyplot as plt
        import rclpy
        from rclpy.node import Node
        from obelisk_sensor_msgs.msg import ObkImage
        from obelisk_py.core.utils.msg import multiarray_to_np, np_to_multiarray
        
        class ImageSubscriber(Node):
        
            def __init__(self):
                super().__init__('image_subscriber')
                self.subscription = self.create_subscription(
                    ObkImage,
                    '/obelisk/dummy/zed2',
                    self.listener_callback,
                    10)
                self.first_time = True
        
            def listener_callback(self, msg):
                self.get_logger().info('Received image data')
                tensor = multiarray_to_np(msg.y)
                if self.first_time:
                    self.first_time = False
                    plt.imshow(tensor[0])
                    plt.show()
                    plt.imshow(tensor[1])
                    plt.show()
        
        def main(args=None):
            rclpy.init(args=args)
            image_subscriber = ImageSubscriber()
            rclpy.spin(image_subscriber)
        
            image_subscriber.destroy_node()
            rclpy.shutdown()
        
        if __name__ == '__main__':
            main()
        ```
- separated out leap and zed code in both the main lib and the ros entrypoints under a `hardware` subdirectory. this more easily allows for conditional builds of hardware packages based on the `setup.sh` script (setup also updated).
- commit to convention for names of some files/targets
    - when writing hardware interfaces, the file should be named something like `<name_of_robot>_interface.h` in c++ or `<name_of_robot>_interface.py` in python
    - when naming targets in `CMakeLists.txt` files, we should use `CamelCase` and the names of the targets for libraries associated with the hardware interface headers should be named `NameOfRobotInterface`. The targets associated with the ROS entrypoint executables should just be called `NameOfRobot`. A similar convention exists for other hardware, like sensors.
    - however, this convention doesn't hold when naming the targets for ROS2, which should still be written in `snake_case`
- removed sensing packages, instead replaced with more specific hardware packages

### Motivation for the PR

- it was discovered that the python node published very slow (about 6 Hz). however, it was determined that this wasn't because the ZED cameras could not be polled at 100Hz, but because simply publishing a raw multiarray onto some topic was extremely slow, motivating the use of ROS' `image_transport` libraries, which can send compressed images using syntax very similar to a publisher
- this feature is only implemented in c++, so it was decided that the ZED node would be written in c++ and not python
- it was discovered that in cpp, we didn't even need to use `image_transport`: simply publishing `ObkImage` message types was already very fast. to test:
    ```
    obk-launch config_file_path=dummy_cpp_zed.yaml device_name=onboard
    ```

### Additional Commentary

- Potentially some flaky behavior when running the zed nodes in a downstream project
    - On the night of July 30, it seemed like the cameras ran ~50Hz slower when called from a local install downstream project in obelisk examples vs in this dev container.
    - On July 31, with some pretty minor changes (like making mutex thinner), I found that the speed was now the same.
    - I've seen online now that `ros2 topic hz` may actually be very slow or inaccurate for high info density topics, so in obelisk_examples, I made a custom counter node in c++ to verify timings (for now, it matches the output of `ros2 topic hz`).
- LEAP hand hardware was verified to be functional.

### Future Issues

- investigate/fix issues with how the `Zoo` target is being created - even if deps are resolved upstream, simply importing `Zoo` doesn't seem to fix it all. For instance, in the zed ROS2 CMakeLists.txt file, we still need to link against the `zed2_sensors_node` target
- suppress warnings from fetched libs (yaml-cpp and zed are big offenders)
- once test suite is filled out in the future, have hardware-conditional tests that aren't run in regular CI
- generalize the zed2 code to all zed cameras (should not be too hard, do this when we incorporate the zed mini)


obelisk_ws/src/obelisk_ros/config/dummy_py_zed.yaml
```diff
@@ -0,0 +1,106 @@
...
+        params_path_pkg: obelisk_ros
+      params_path: config/dummy_py_zed_params.yaml
+      # callback_groups:
+      publishers:
+        - ros_parameter: pub_img_setting
+          topic: /obelisk/dummy/zed2
+          msg_type: ObkImage
+          history_depth: 10
+          callback_group: None
+          non_obelisk: False
```
alberthli, line 64: [note] this is the only difference between this file and `dummy_py.yaml`

obelisk/cpp/zoo/CMakeLists.txt
```diff
@@ -22,7 +22,20 @@ add_library(Zoo INTERFACE)
 add_library(Obelisk::Zoo ALIAS Zoo)
 
 target_include_directories(Zoo INTERFACE include)
-target_include_directories(Zoo INTERFACE robot)
+
+# conditionally add leap subdirectory if OBELISK_BUILD_LEAP is set
```
Zolkin1, line 26: Maybe we should add a `message` line that says which options are on/off. I find that down the line this is useful to have so you can be sure certain flags are set and correct things are building.

obelisk/cpp/zoo/hardware/sensing/zed/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+FetchContent_MakeAvailable(yaml-cpp)
+
+# Disable warnings for yaml-cpp
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    target_compile_options(yaml-cpp PRIVATE -w -Wno-unused-parameter)
+elseif(MSVC)
+    target_compile_options(yaml-cpp PRIVATE /W0 /wd4100)
+endif()
+
+# ----------- Making the executables ----------- #
```
Zolkin1, line 55: nit: this would be "Making the Libraries"

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
```diff
@@ -0,0 +1,420 @@
...
+        pub_period_                 = this->get_parameter("pub_period").as_double();
+        std::string params_path_pkg = this->get_parameter("params_path_pkg").as_string();
+        std::string _params_path    = this->get_parameter("params_path").as_string();
+
+        std::string params_path;
+        if (_params_path[0] == '/') {
+            params_path = _params_path;
+        } else if (params_path_pkg.empty()) {
+            std::string err_msg = "No package provided for ZED2 camera parameters!";
+            RCLCPP_ERROR(this->get_logger(), err_msg.c_str());
```
Zolkin1, line 47: There is technically nothing wrong with this, but I think its cleaner to just use `RCLCPP_ERROR_STREAM` then you can just easily put the message in as a normal stream. I.e. you don't have to make a string in the line above. Tbh even with the non-stream option I doubt you need that string, and it can prob be just written in the error line.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
```diff
@@ -0,0 +1,420 @@
...
+        // Declare parameters
+        this->declare_parameter("params_path_pkg", "");
+        this->declare_parameter("pub_period", 0.0);
+
+        // Get parameter values
+        pub_period_                 = this->get_parameter("pub_period").as_double();
+        std::string params_path_pkg = this->get_parameter("params_path_pkg").as_string();
+        std::string _params_path    = this->get_parameter("params_path").as_string();
+
+        std::string params_path;
```
Zolkin1, line 42: I've been trying to use std::filesystem::path for paths. Tbh its effectively a string, but it has the semantic meaning. Also you can use functions like `std::filesystem::exists` with it to check if the path exists.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
```diff
@@ -0,0 +1,420 @@
...
+
+    // Callback group shared by all camera callbacks
+    rclcpp::CallbackGroup::SharedPtr cam_cbg_;
+
+    // Camera parameter maps
+    std::map<int, std::map<std::string, std::string>> cam_param_dicts_;
+    std::map<int, sl::Camera> cams_;
+    std::map<int, sl::Mat> cam_images_;
+    std::map<int, sl::RuntimeParameters> cam_rt_params_;
+    std::map<int, Eigen::Tensor<uint8_t, 3>> frames_;
```
Zolkin1, line 179: did you want to change these to vectors, or is there a reason they are maps?

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
```diff
@@ -0,0 +1,420 @@
...
+
+    /**
+     * @brief Set camera parameters from a YAML configuration file.
+     *
+     * @param params_path Path to the YAML configuration file.
+     */
+    void set_camera_params(const std::string& params_path) {
+        // Check if the file has a .yaml or .yml extension
+        if (params_path.substr(params_path.find_last_of(".") + 1) != "yaml" &&
+            params_path.substr(params_path.find_last_of(".") + 1) != "yml") {
```
Zolkin1, line 199: If you use std::filesystem::path I think there are helper functions to get extensions that make this cleaner.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
```diff
@@ -0,0 +1,420 @@
...
+            std::string err_msg = "Invalid parameters file: " + params_path +
+                                  "! ObeliskZed2Sensors expects a YAML file with extension .yaml or .yml.";
+            RCLCPP_ERROR(this->get_logger(), err_msg.c_str());
```
Zolkin1, line 202: Once again with the extra string, probably won'r mark anymore if there are more.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
```diff
@@ -0,0 +1,420 @@
...
+                std::string err_msg = "Invalid side for camera " + std::to_string(cam_index) + " in " + params_path +
+                                      "! Must be one of 'left', 'right'.";
+                RCLCPP_ERROR(this->get_logger(), err_msg.c_str());
+                throw std::runtime_error(err_msg);
+            }
+            cam_param_dicts_[cam_index]["side"] = side;
+        }
+
+        // Check that all cameras have the same resolution, fps, and depth setting
+        std::set<std::string> resolutions, fpss, depths;
```
Zolkin1, line 316: Not sure why this is a std::set and not another container. I've never used set before, but it looks like its useful for sorting and idk if you are taking advantage of that.

obelisk/cpp/zoo/CMakeLists.txt
```diff
@@ -22,7 +22,20 @@ add_library(Zoo INTERFACE)
...
+if(OBELISK_BUILD_LEAP)
+  message(STATUS "Configuring Leap Hand Interface")
+  add_subdirectory(hardware/robots/leap)
+  target_include_directories(Zoo INTERFACE hardware/robots/leap)
+endif()
+
+# conditionally add zed subdirectory if OBELISK_BUILD_ZED is set
+if(OBELISK_BUILD_ZED)
```
Zolkin1, line 34: If these are environment variables, dont you need `$ENV{VAR}` to access them? https://cmake.org/cmake/help/latest/variable/ENV.html

Zolkin1 (approved):
Overall looks good. Left a number of relatively minor comments.

obelisk/cpp/obelisk_cpp/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ FetchContent_Declare(
   URL https://github.com/google-deepmind/mujoco/releases/download/${MUJOCO_VERSION}/mujoco-${MUJOCO_VERSION}-linux-${COMP_ARCH}.tar.gz
 )
 
-FetchContent_Populate(mujoco)
+FetchContent_MakeAvailable(mujoco)
```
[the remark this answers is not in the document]
alberthli: This actually came up when I accidentally did circular Obelisk imports. One thing that I kept seeing was that it was complaining that mujoco was already built (from the first part of the circular call), and that went away when I replaced this. It doesn't complain now.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: so here's the logic: I originally had these be strings (camera names), but then I realized that there was no way to guarantee their order if just indexed by the string name. Thus, I changed the names to integers. In loops over these maps, the values are drawn by order of their keys, which means that technically, the names don't have to be things like 0, 1, ... It can be something like 51, 73, ..., which is stupid, but doesn't impose any restrictions on the end user filling out the parameter yaml file.

If you feel strongly about this, we can change all these maps to vectors and I can change the yaml formatting too, though... I might be able to make it a list so there aren't even keys associated with the camera settings.

obelisk/cpp/zoo/CMakeLists.txt
alberthli: resolved: now, i also added an `OR` condition with the `$ENV` syntax, since per slack, these can be either a cmake arg or env var due to redundancies in obelisk, and I'm not sure how to properly distinguish between them.

obelisk/cpp/zoo/hardware/sensing/zed/CMakeLists.txt
alberthli: resolved

obelisk/cpp/utils/include/msg_conversions.h
```diff
@@ -46,7 +46,20 @@ namespace obelisk::utils::msgs {
             sizes.at(i) = msg.layout.dim.at(i).size;
         }
 
-        auto tensor = internal::CreateTensor<double, Size>(data, sizes, std::make_index_sequence<Size>{});
```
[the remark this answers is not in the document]
alberthli: resolved.

obelisk/cpp/zoo/CMakeLists.txt
alberthli: don't i have one right underneath each if statement already?
```
message(STATUS "Configuring Leap Hand Interface")
```
or are you saying we should always print even if the if statement is not true?

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: The main reason I do this is because I not only want to display this error, I also want to pass the error message into a runtime error that I throw. Therefore, it's cleaner for me to define the error message string and then pass to both.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: I'll resolve this conversation - let's consolidate to the other thread.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: This was a holdover that got translated from some python logic (which tbf also didn't need to bust out sets). I've refactored it.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: resolved.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: resolved.

obelisk/cpp/zoo/CMakeLists.txt
alberthli: resolved in person.

obelisk/cpp/zoo/hardware/sensing/zed/zed2_sensors_interface.h
alberthli: as per f2f: changed all maps to std::vectors and refactored code to work with vectors. to facilitate this change, the camera parameter yaml file is now a list instead of a dict with int keys.

PR #117, opened 2024-08-28
Fixed simulation timing issues by making the simulator sim in real time.

Zolkin1:
Changed the simulator to simulate in real time. Moved rendering to a separate thread.

### DIFF
--- a/obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
+++ b/obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
@@ -164,6 +164,62 @@ namespace obelisk {
 
             data_ = mj_makeData(model_);
 
+            // Create the rendering thread
+            std::thread rendering_thread(std::bind(&ObeliskMujocoRobot::SimRender, this));
+
+            mujoco_setup_ = true;
+
+            while (!activation_complete_) {
+                std::this_thread::sleep_for(std::chrono::milliseconds(3));
+            }
+            RCLCPP_INFO_STREAM(this->get_logger(), "Starting Mujoco simulation loop.");
+
+            // Set the initial condition based on a keyframe
+            RCLCPP_INFO_STREAM(this->get_logger(), "Found " << this->model_->nkey << " Mujoco keyframes.");
+            for (int i = 0; i < this->model_->nkey; i++) {
+                std::string potential_keyframe(this->model_->names + this->model_->name_keyadr[i]);
+                if (potential_keyframe == this->get_parameter("ic_keyframe").as_string()) {
+                    RCLCPP_INFO_STREAM(this->get_logger(),
+                                       "Setting initial condition to keyframe: " << potential_keyframe);
+                    mju_copy(this->data_->qpos, &this->model_->key_qpos[i * this->model_->nq], this->model_->nq);
+                    mju_copy(this->data_->qvel, &this->model_->key_qvel[i * this->model_->nv], this->model_->nv);
+                    this->data_->time = this->model_->key_time[i];
+
+                    mju_copy(this->data_->ctrl, &this->model_->key_ctrl[i * this->model_->nu], this->model_->nu);
+                }
+            }
+
+            while (!this->stop_thread_) {
+                auto start_time = this->now();
+                {
+                    // Get the data from the shared vector safely
+                    std::lock_guard<std::mutex> lock(shared_data_mut_);
+                    for (int i = 0; i < nu_; i++) {
+                        data_->ctrl[i] = shared_data_.at(i);
+                    }
+                }
+
+                {
+                    std::lock_guard<std::mutex> lock(sensor_data_mut_);
+                    mj_step(model_, data_);
+                }
+
+                // Thread sleeping won't be precise enough for this, so instead we will spin lock the thread. This is
+                // inefficient, but accurate.
+                while ((this->now() - start_time).nanoseconds() < time_step_ * 1e9) {
+                }
+            }
+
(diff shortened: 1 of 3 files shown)


obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -164,6 +164,62 @@ namespace obelisk {
 
             data_ = mj_makeData(model_);
 
+            // Create the rendering thread
+            std::thread rendering_thread(std::bind(&ObeliskMujocoRobot::SimRender, this));
```
alberthli, line 168: would it make sense to initialize this thread elsewhere like in configuration? that way you can also manage it from the lifecycle callbacks

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -164,6 +164,62 @@ namespace obelisk {
...
+
+            mujoco_setup_ = true;
+
+            while (!activation_complete_) {
+                std::this_thread::sleep_for(std::chrono::milliseconds(3));
```
alberthli, line 173: is this always consistently enough time? i'd consider bumping up the delay here just to really make sure it's safely compatible with whatever system this is run on that's slower than your machine

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -164,6 +164,62 @@ namespace obelisk {
...
+                if (potential_keyframe == this->get_parameter("ic_keyframe").as_string()) {
+                    RCLCPP_INFO_STREAM(this->get_logger(),
+                                       "Setting initial condition to keyframe: " << potential_keyframe);
+                    mju_copy(this->data_->qpos, &this->model_->key_qpos[i * this->model_->nq], this->model_->nq);
+                    mju_copy(this->data_->qvel, &this->model_->key_qvel[i * this->model_->nv], this->model_->nv);
+                    this->data_->time = this->model_->key_time[i];
+
+                    mju_copy(this->data_->ctrl, &this->model_->key_ctrl[i * this->model_->nu], this->model_->nu);
+                }
+            }
```
alberthli, line 190: so in the case a keyframe is not found, it sets everything to 0s (and identity quats), right?

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -164,6 +164,62 @@ namespace obelisk {
...
+                while ((this->now() - start_time).nanoseconds() < time_step_ * 1e9) {
+                }
+            }
+
+            RCLCPP_WARN_STREAM(this->get_logger(), "Cleaning up simulation data and model...");
+            // free MuJoCo model and data
+            mj_deleteData(data_);
+            mj_deleteModel(model_);
+
+            // rendering_thread.join();
```
alberthli, line 218: nit: delete this before merge, though you might want to join the thread in cleanup and/or shutdown explicitly, which would require moving the thread initialization out

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
```diff
@@ -256,26 +280,15 @@ namespace obelisk {
...
                     glfwPollEvents();
 
-                    auto end_time   = std::chrono::steady_clock::now();
-                    auto total_time = std::chrono::duration_cast<std::chrono::microseconds>(end_time - start_time);
-                    const int des_time_us = static_cast<int>((num_steps_per_viz_ * time_step_) * 1e6);
-                    if (total_time.count() < des_time_us) {
-                        const int time_diff_us = des_time_us - total_time.count();
-                        std::this_thread::sleep_for(std::chrono::microseconds(time_diff_us));
-                    }
+                    std::this_thread::sleep_for(std::chrono::milliseconds(17)); // 58Hz render rate at most
```
alberthli, line 283: is this just specifically tuned for your machine? we should have something in place that can ensure settings that are robust for many machines or warn if not if we're merging into main

obelisk_ws/src/obelisk_ros/config/dummy_cpp.yaml
```diff
@@ -65,7 +65,8 @@ onboard:
       executable: obelisk_mujoco_robot
       params:
         ic_keyframe: ic
-      # callback_groups:
+      callback_groups:
+        rendering_cbg: MutuallyExclusiveCallbackGroup
```
alberthli, line 69: do you use this anywhere? i don't see it

alberthli (approved):
Approved, but some lingering questions

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
Zolkin1: There is nothing important happening here actually, I am just waiting for the rest of the configuration in other threads to complete, so the timing is not very important.

obelisk/cpp/obelisk_cpp/include/obelisk_mujoco_sim_robot.h
Zolkin1: This should be machine independent. But that being said, as of now there is no option to adjust the frame rate for the rendering, but I think rendering at around 60Hz is pretty reasonable. I guess maybe there is a concern that the machine doesn't have the capability to render at 60Hz, but honestly I don't really think thats a big issue.

PR #126, opened 2024-11-07
Mujoco sim upgrades

Zolkin1:
Updated some features in the Mujoco simulation:
- Removed the need to specify `nu` in the yaml, instead we now read that from the Mujoco model. This was annoying because if you forgot to update the yaml it was hard to figure out what the issue was.
- Removed the need to specify the sim timestep in the yaml, instead we now read that from the Mujoco model. In the future we can add back the option to overwrite the Mujoco time step with what the user puts in the yaml, but for now that seems unnecessary.

Updated the docs to be more accurate.

### DIFF
--- a/docs/source/using_obelisk.rst
+++ b/docs/source/using_obelisk.rst
@@ -311,27 +311,20 @@ Obelisk nodes can be easily configured via a Obelisk configuration (yaml) file. 
         # callback_groups:
         publishers:
           - ros_parameter: pub_ctrl_setting
-            # key: pub1
             topic: /obelisk/dummy/ctrl
             msg_type: PositionSetpoint
             history_depth: 10
             callback_group: None
-            non_obelisk: False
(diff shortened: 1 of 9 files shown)


docs/source/using_obelisk.rst
```diff
@@ -311,82 +311,79 @@ Obelisk nodes can be easily configured via a Obelisk configuration (yaml) file.
...
             callback_group: None
-            # callback_key: timer_callback1
     estimation:
       - pkg: obelisk_estimation_cpp
         executable: jointencoders_passthrough_estimator
         # callback_groups:
         publishers:
           - ros_parameter: pub_est_setting
             # key: pub1
-            topic: /obelisk/dummy/est
```
alberthli, line 342: did you mean to delete this?

alberthli (approved):
lgtm

PR #136, opened 2025-06-10
Update Setup

alberthli:
Minimal updates to the Obelisk setup in preparation for writing the FR3 hardware interface


README.md
```diff
@@ -3,13 +3,46 @@ This repository hosts generic interfaces for controlling robots with ROS2.
 
 [Docs](https://caltech-amber.github.io/obelisk/)
 
-## Setup
+# Quickstart
+
+This section contains Quickstart information for members of the AMBER Lab.
+
+## FR3
```
Zolkin1, line 10: I think we should put this type of stuff in the online docs and keep the readme as slim as possible.

alberthli: as per slack, i opened instead issue #140

PR #163, opened 2026-06-01
refactor scans to use multi-scan

wdc3iii:
mj_multiray should provide performance improvement over looping mj_ray by amortizing setup costs. Only works for raycasts which share a common origin.


Zolkin1 (approved):
lgtm

